PR #890, opened 2024-12-12
PYR1-970 add miles and kilometers to measure distance tool modal

drewverlee:
## Purpose
Given that most distances calculated are over a thousand meters, it's easier to read in kilometers than in meters. However, our users don't use meters regularly because they are from the U.S., so we should prefer showing the user's miles over kilometers. 

## Related Issues
Closes PYR1-970

## Submission Checklist

- [ i think so? Hard to say] Code passes linter rules (`clj-kondo --lint src`)
- [ runs just fine] Feature(s) work when compiled (`clojure -M:compile-cljs`)
- [not sure how to tell if there are new ones, i doubt it ] No new reflection warnings (`clojure -M:check-reflection`)
- [ no need] For any large features/rearchitecting of the codebase, the relevant `docs` were updated (e.g. updating `docs/pyrecast-database.md` when adding a new DB table)

#### Module Impacted
I'm not sure what module was impacted.

#### Role
I assume all roles can see this chnage

#### Steps
measure a distance using the sidebar tool 

#### Desired Outcome
The measured distance should be displayed in miles and kilometers now instead of meters.

## Screenshots

Before, when it was just meters

image

After, it shows miles and kilometers.  (notice i also increased the font size to draw attention to the distance, we can kick this back if we want)
image

### DIFF
--- a/src/cljs/pyregence/components/map_controls/measure_tool.cljs
+++ b/src/cljs/pyregence/components/map_controls/measure_tool.cljs
@@ -58,9 +58,6 @@
                               :justify-content "center"
                               :margin          ".4rem"
                               :font-size       "1.2rem"}}
-             ;; Below we convert distance-between-points, which is in meters,
-             ;; to both miles and kilometers which is ideal for larger distances
-             ;; and for the majority of our U.S based users.
              (let [[mi km] (->>  [[@distance-between-points "miles" #(* 0.00062137 %)]
                                   [@distance-between-points "kilometers" #(/ % 1000)]]
                                  (map (fn [[m l f]] (str (cl-format nil "~,1f" (f m)) " " l))))]
--- a/src/cljs/pyregence/components/map_controls/measure_tool.cljs
+++ b/src/cljs/pyregence/components/map_controls/measure_tool.cljs
@@ -59,8 +59,8 @@
                               :font-size       "1.2rem"}}
              (let [format (fn [number label] (str (cl-format nil "~,1f" number) " " label))
                    meters  @distance-between-points
-                   miles (* 0.00062137 meters)
-                   kilometers (/ meters 1000)
+                   miles (* meters 0.00062137)
+                   kilometers (* meters 0.001)
                    mi (format miles "miles")
                    km (format kilometers "kilometers")]
                (str mi " (" km ")" ))])
--- a/src/cljs/pyregence/components/map_controls/measure_tool.cljs
+++ b/src/cljs/pyregence/components/map_controls/measure_tool.cljs
@@ -60,7 +60,8 @@
                               :font-size       "1.2rem"
                               :justify-content "center"
                               :margin          ".4rem"}}
-             (str (meters->miles @distance-between-points) " (" (meters->kilometers @distance-between-points) ")")])
+             (str (meters->miles @distance-between-points)
+                  " (" (meters->kilometers @distance-between-points) ")")])
           [:div {:style {:display        "flex"
                          :flex-direction "column"
                          :margin-top     ".2rem"}}
--- a/src/cljs/pyregence/components/map_controls/measure_tool.cljs
+++ b/src/cljs/pyregence/components/map_controls/measure_tool.cljs
@@ -58,6 +58,9 @@
                               :justify-content "center"
                               :margin          ".4rem"
                               :font-size       "1.2rem"}}
+             ;; Below we convert distance-between-points, which is in meters,
+             ;; to both miles and kilometers which is ideal for larger distances
+             ;; and for the majority of our U.S based users.
              (->>  [[@distance-between-points "miles" #(* 0.00062137 %)]
                     [@distance-between-points "kilometers" #(/ % 1000)]]
                    (map (fn [[m l f]] (str (cl-format nil "~,1f" (f m)) " " l)))
--- a/src/cljs/pyregence/components/map_controls/measure_tool.cljs
+++ b/src/cljs/pyregence/components/map_controls/measure_tool.cljs
@@ -63,7 +63,7 @@
                    kilometers (* meters 0.001)
                    mi (format miles "miles")
                    km (format kilometers "kilometers")]
-               (str mi " (" km ")" ))])
+               (str mi " (" km ")"))])
           [:div {:style {:display        "flex"
                          :flex-direction "column"
                          :margin-top     ".2rem"}}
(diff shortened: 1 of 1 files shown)


src/cljs/pyregence/components/map_controls/measure_tool.cljs
```diff
@@ -52,13 +53,21 @@
...
           [lon-lat-position "Point Two Location" (if @point-two @point-two [0 0])]
+          (when (> @distance-between-points 0)
+            [:strong {:style {:display         "flex"
+                              :justify-content "center"
+                              :margin          ".4rem"
+                              :font-size       "1.2rem"}}
+             ;; Below we convert distance-between-points, which is in meters,
+             ;; to both miles and kilometers which is ideal for larger distances
+             ;; and for the majority of our U.S based users.
+             (->>  [[@distance-between-points "miles" #(* 0.00062137 %)]
```
drewverlee, line 64: is `#(* 0.00062137 %)` the best way to convert meters to miles?

drewverlee:
I want feedback on the code used in the conversions from meters to kilometers and miles. Here is the code for meters to miles `#(* 0.00062137 %)` here is the code for meters to kilometers  `#(/ % 1000)`.

Is this an existing library for this? Are these conversions as accurate as they need to be?

drewverlee:
Sif made a good suggestion on using parens instead of a forward slash.

drewverlee:
updated to use parens:

image

src/cljs/pyregence/components/map_controls/measure_tool.cljs
```diff
@@ -52,13 +52,21 @@
...
+                              :justify-content "center"
+                              :margin          ".4rem"
+                              :font-size       "1.2rem"}}
+             (let [format (fn [number label] (str (cl-format nil "~,1f" number) " " label))
+                   meters  @distance-between-points
+                   miles (* 0.00062137 meters)
+                   kilometers (/ meters 1000)
+                   mi (format miles "miles")
+                   km (format kilometers "kilometers")]
+               (str mi " (" km ")" ))])
```
Oliver-BE, line 66: Remove extra space:
```suggestion
               (str mi " (" km ")"))])
```

src/cljs/pyregence/components/map_controls/measure_tool.cljs
```diff
@@ -52,13 +52,21 @@
...
-                         :margin-top      ".2rem"}}
-           (when (> @distance-between-points 0)
-             [:div {:style {:height 30 :margin ".2rem"}}
-              [:label {:style ($/padding "1px" :1)}
-               (str (cl-format nil "~,4f" @distance-between-points) " meters")]])
+          (when (> @distance-between-points 0)
+            [:strong {:style {:display         "flex"
+                              :justify-content "center"
+                              :margin          ".4rem"
+                              :font-size       "1.2rem"}}
```
Oliver-BE, line 59: Can you organize these style tags in alphabetical order? We try to do this throughout the code base for consistency :)

src/cljs/pyregence/components/map_controls/measure_tool.cljs
```diff
@@ -52,13 +52,21 @@
...
+             (let [format (fn [number label] (str (cl-format nil "~,1f" number) " " label))
+                   meters  @distance-between-points
+                   miles (* 0.00062137 meters)
+                   kilometers (/ meters 1000)
+                   mi (format miles "miles")
+                   km (format kilometers "kilometers")]
```
Oliver-BE: Another stylistic thing that we follow in Pyrecast: we align columns in `let` bindings like so:
```suggestion
             (let [format     (fn [number label] (str (cl-format nil "~,1f" number) " " label))
                   meters     @distance-between-points
                   miles      (* 0.00062137 meters)
                   kilometers (/ meters 1000)
                   mi         (format miles "miles")
                   km         (format kilometers "kilometers")]
```

Oliver-BE: Also, it's not necessary, but if you wanted to add the meters to miles and meters to kilometers conversions to a util file, the right one would be `src/cljs/pyregence/utils/number_utils.cljs`.

src/cljs/pyregence/components/map_controls/measure_tool.cljs
```diff
@@ -52,13 +52,21 @@
...
+            [:strong {:style {:display         "flex"
+                              :justify-content "center"
+                              :margin          ".4rem"
+                              :font-size       "1.2rem"}}
+             (let [format (fn [number label] (str (cl-format nil "~,1f" number) " " label))
+                   meters  @distance-between-points
+                   miles (* meters 0.00062137)
+                   kilometers (* meters 0.001)
+                   mi (format miles "miles")
+                   km (format kilometers "kilometers")]
```
Oliver-BE, line 65: We typically try to keep `let` bindings at the top-level. Can you move these to within the outer `r/with-let`?

Also, another stylistic thing we do is align columns within bindings like so:
```suggestion
             (let [format     (fn [number label] (str (cl-format nil "~,1f" number) " " label))
                   meters     @distance-between-points
                   miles      (* meters 0.00062137)
                   kilometers (* meters 0.001)
                   mi         (format miles "miles")
                   km         (format kilometers "kilometers")]
```

Oliver-BE (commented):
Nice work! Just a couple of small comments. Also,  can you make sure your PRs start with the Jira code so that our Jira Automation task can pick it up properly (e.g. "PYR1-970" instead of "Pyr1 970")?

danielhvs:
> David made a good suggestion on using parens instead of a forward slash.

You can call me David or Sif if you'd like. LOL

Oliver-BE:
> > David made a good suggestion on using parens instead of a forward slash.
> 
> You can call me David or Sif if you like. LOL

:man_facepalming: sorry!! Fixed it

src/cljs/pyregence/components/map_controls/measure_tool.cljs
drewverlee-sig: > We typically try to keep let bindings at the top-level. Can you move these to within the outer r/with-let?

Kind of, I don't think we can bind mi and km inside r/with-let as is done in the commit here because r/with-let only evals once, which would lock both of those in as 0. 

We can bind a function in the r/with-let which can take the meters/distance-between-points. That's what the latest [commit ](https://github.com/pyregence/pyregence/pull/890/commits/fbc48f3796264d8589ac2616dc0dbebf80b32c24) does, i'm not sure that's more clear.  I think either approach is fine though I prefer tighter scopes, less chance of confusion.

PR #891, opened 2024-12-18
PYR1-1014 remove camera icons with inactive feeds as icons on the map

drewverlee:
Remove inactive cameras

Inactive means cameras that haven't updated in four hours. To discover that, we had to reformat the non-California cameras so that they had a valid UTC timestamp, just like the California cameras did. We have very high confidence
that they are UTC timestamps. 

Now that we're removing inactive cameras on the server side, so they never reach the front end, the logic on the fe dealing with that can be removed. 

To see a visual example of things before and after removing the inactive cameras, see the picture below, which shows two browsers: one on top of the other, where the first is on pyrecast.org, and so currently doesn't have this commit's change (still shows inactive cameras) and the bottom, a local running env, does have this commits change and so doesn't show the selected  inactive cameras. 

You can see that the orange (selected) camera in the top browser is inactive ("not refreshed for 11257 hours), but that camera doesn't show up in the bottom browser, which fulfills the requirement of the ticket. 

image


src/cljs/pyregence/components/map_controls/camera_tool.cljs
```diff
@@ -187,21 +173,12 @@
...
-                                                                 (u-time/alert-wf-camera-time->js-date %)
-                                                                 (js/Date. %))
-                                                               (u-time/get-time-difference %)
-                                                               (u-time/ms->hr %)))
-                                      (when (> 4 @camera-age)
-                                        (reset! image-src (<! image-chan))
-                                        (reset! exit-chan
-                                                (u-async/refresh-on-interval! #(go (reset! image-src (<! (get-camera-image-chan @active-camera))))
-                                                                              60000)))))))
+                                      (reset! image-src (<! image-chan))
```
drewverlee, line 178: A second pair of eyes on this change would be nice as I haven't reviewed it thoroughly.

Oliver-BE:
Nice work! A few small comments:

1. We can remove the following imports in `camera_tool.cljs` as they aren't being used anymore:
```clojure
            [pyregence.utils.number-utils                  :as u-num]
            [pyregence.utils.time-utils                    :as u-time]
```
2. Can you spell-check and add caps/periods to your docstrings in `cameras.clj`? (e.g. "returns true if the utc timestamp is older then four hours" -> "Returns true if the UTC timestamp is older than four hours.")
3. Remove extra line on line 206 in `cameras.clj`.

drewverlee:
@Oliver-BE Thanks for the review. I made the suggested changes (removing extra lines, grammar changes, and unused imports). Let me know if anything else needs to be done.

PR #893, opened 2024-12-24
PYR1-1016 logged in users map should stay where they are when switching between tabs

drewverlee:
This change disables auto-zooming when on the Risk or PSPS tab. If a Utility Company wants to re-enable auto-zooming for a given Risk Ignition Pattern, we can do that by updating the database like so:

```
UPDATE public.organization_layers
  SET layer_config='{:auto-zoom? true <TODO rest of config ommited>}'
  WHERE org_layer_uid=<TODO>;
```

Note some` < values >` have been omitted.

### DIFF
--- a/src/cljs/pyregence/config.cljs
+++ b/src/cljs/pyregence/config.cljs
@@ -410,7 +410,6 @@
                                                                                 :units        "Ignitions/line-mi/hr"
                                                                                 :disabled-for #{:all :tlines}}}}
                                     :pattern    {:opt-label  "Ignition Pattern"
-                                                 :auto-zoom? true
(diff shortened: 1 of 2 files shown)


Oliver-BE:
Let's set `:auto-zoom?` to true for the `:all` and `:tlines` options. This way, normal users will still have the risk tab auto zoom to those layers. Utility company users won't have the risk tab auto zoom to their layers because you removed the `:auto-zoom? true` directly on the `:pattern` drop down. I don't think we need to apply the changefile you provided on Jira -- we don't want to `:auto-zoom?` on this risk layers for utility companies.

PR #894, opened 2024-12-26
PYR1-991 add default ignition pattern using default option

drewverlee:
This change adds the ability to select a default Risk Ignition Pattern by updating the Pyregence database as shown below:

```
INSERT INTO public.organization_layers (organization_rid,layer_path,layer_config) VALUES
	 (<TODO>,'[:fire-risk :params :pattern]','{:default-option <TODO>}');
```

Where the `:default-option` should be one of the existing ignition patterns already in the database's layer_path e.g the `:utilitycompany` at the end of the layer path:`[:fire-risk :params :pattern :options :utilitycompany]`. 

Beyond the above SQL insert, two changes to the code are necessary. 

First, it allows for the full near_term_forecasting_capabilities layer options to be built from more than one layer_config row. This is accomplished by merging layer configs rather than associating/replacing them. When I checked the prod data given, the only difference in data this produced, given the example SQL Insert above, was the desired one. 

Secondly, this change requires a modification to the filtering of layer_configs because, as verified in this [commit ](https://github.com/pyregence/pyregence/commit/ad7be09d72682a6bc4ebad2a77531044cb4e7d8b) the necessary layer config (added in the sql above) is invalid according to the current layer_config spec:

```
  (s/valid? ::layer-config {:default-option :UtilityCompany})
  ;; => false

```

This change to filter requires some consideration and is the primary reason I recommended going with using the :selected? key in 
this[ pull request.](https://github.com/pyregence/pyregence/pull/892). 

All options are on the table, and this PR currently contains the easiest one (as I'm not sure what the simplest one is!). As a starting point, it's unclear to me what the purpose of the validation is on read instead of on write, where it could be used to give the writer feedback so they could correct the validation problems and re-submit the correctly validated data.

If we want to keep the validations on read and the way the state is currently constructed, we could try to validate the reduced result of merging rather than the pieces merged in. I can explore that option if it sounds promising, though it's worth keeping in mind that the function result, before this PR, was nothing, and after the PR, the validations cause an issue (which is why they were removed). 

Let me know if you have further thoughts!

Here are the before and after photos. Before:

image

After, with changes and sql applied (some layers aren't showing but thats a local dev issue that doesn't effect the PR as far as i know)

image


src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -34,25 +33,6 @@
...
-(s/def ::opt-label     string?)
-(s/def ::units         string?)
-(s/def ::z-index       int?)
-(s/def ::layer-config  (s/keys :req-un [::opt-label (or ::filter ::filter-set)]
-                               :opt-un [::clear-point? ::geoserver-key ::units ::z-index]))
-(s/def ::layer-path    (s/and (s/coll-of keyword? :kind vector? :min-count 2)
-                              (s/cat :forecast #{:fuels :fire-weather :fire-risk :active-fire :psps-zonal}
-                                     :second   #(or (= % :params)
-                                                    (= % :underlays))
-                                     :etc      (s/+ keyword?))))
```
Oliver-BE, line 54: Why don't we just add another optional spec option to the `::layer-path` spec? We can add something like:
```clojure
(s/def ::default-option keyword?)
```
and then update `::layer-config` to include `::default-option` in the `:opt-un` section.

Oliver-BE: Actually, ignore this. I think I've come up with an easier way to make this update that doesn't involve updating the DB at all. I'll push a commit with my thoughts shortly.

Oliver-BE:
@drewverlee I pushed a [new commit](https://github.com/pyregence/pyregence/pull/894/commits/df4c574489a029fc22faa938c9c52cfc5b9317e0) that solves the issue of setting the default ignition pattern option without needing to update the DB. Let me know what you think!

PR #896, opened 2024-12-31
PYR1-975 add toast message when there are no active fires

drewverlee:
Toast "No active fires." when there are no active fires and it's the first time the active fire tab has been visited.

### DIFF
--- a/src/cljs/pyregence/components/forecast_tabs.cljs
+++ b/src/cljs/pyregence/components/forecast_tabs.cljs
@@ -1,8 +1,5 @@
 (ns pyregence.components.forecast-tabs
-  (:require
-   [pyregence.components.common    :refer [tool-tip-wrapper]]
-   [pyregence.components.messaging :refer [toast-message!]]
-   [pyregence.state                :as !]))
+  (:require [pyregence.components.common :refer [tool-tip-wrapper]]))
 
 (defn- $forecast-label [selected?]
   (merge
@@ -25,14 +22,6 @@
               hover-text
               :top
               [:label {:style    ($forecast-label (= current-forecast key))
-                       :on-click
-                       (fn []
-                         (when (= key :active-fire)
-                           (when (and
-                                  (zero? @!/active-fire-count)
-                                  (zero? @!/active-fire-tab-click-count))
-                             (toast-message! "No active fires."))
-                           (swap! !/active-fire-tab-click-count inc))
-                         (select-forecast! key))}
+                       :on-click #(select-forecast! key)}
                opt-label]]))
          capabilities))])
--- a/src/cljs/pyregence/components/forecast_tabs.cljs
+++ b/src/cljs/pyregence/components/forecast_tabs.cljs
@@ -1,5 +1,8 @@
 (ns pyregence.components.forecast-tabs
-  (:require [pyregence.components.common :refer [tool-tip-wrapper]]))
+  (:require
+   [pyregence.components.common    :refer [tool-tip-wrapper]]
+   [pyregence.components.messaging :refer [toast-message!]]
+   [pyregence.state                :as !]))
 
 (defn- $forecast-label [selected?]
   (merge
@@ -22,6 +25,17 @@
               hover-text
               :top
               [:label {:style    ($forecast-label (= current-forecast key))
-                       :on-click #(select-forecast! key)}
+                       :on-click
+                       (fn []
+                         (when (and
+                                (= key :active-fire)
+                                (zero? (->> @!/processed-params
+                                            :fire-name
+                                            :options
+                                            count
+                                            ;; dec to remove "active fires"
+                                            dec)))
+                           (toast-message! "No active fires"))
+                         (select-forecast! key))}
                opt-label]]))
          capabilities))])
--- a/src/cljs/pyregence/components/forecast_tabs.cljs
+++ b/src/cljs/pyregence/components/forecast_tabs.cljs
@@ -27,14 +27,7 @@
               [:label {:style    ($forecast-label (= current-forecast key))
                        :on-click
                        (fn []
-                         (when (and
-                                (= key :active-fire)
-                                (zero? (->> @!/processed-params
-                                            :fire-name
-                                            :options
-                                            count
-                                            ;; dec to remove "active fires"
-                                            dec)))
+                         (when (and (= key :active-fire) @!/show-no-active-fires?)
                            (toast-message! "No active fires"))
                          (select-forecast! key))}
                opt-label]]))
--- a/src/cljs/pyregence/components/forecast_tabs.cljs
+++ b/src/cljs/pyregence/components/forecast_tabs.cljs
@@ -28,7 +28,7 @@
                        :on-click
                        (fn []
                          (when (and (= key :active-fire) @!/show-no-active-fires?)
-                           (toast-message! "No active fires"))
+                           (toast-message! "No active fires."))
                          (select-forecast! key))}
                opt-label]]))
          capabilities))])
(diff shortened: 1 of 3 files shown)


src/cljs/pyregence/components/forecast_tabs.cljs
```diff
@@ -22,6 +25,14 @@
...
-                       :on-click #(select-forecast! key)}
+                       :on-click
+                       (fn []
+                         (when (= key :active-fire)
+                           (when (and
+                                  (zero? @!/active-fire-count)
+                                  (zero? @!/active-fire-tab-click-count))
+                             (toast-message! "No active fires."))
+                           (swap! !/active-fire-tab-click-count inc))
+                         (select-forecast! key))}
```
danielhvs, line 36: What happens in `select-forecast!` if there is no active fires really?
Wouldn't be better to avoid calling `select-forecast!` function in such a case?

danielhvs (approved):
Just a wondering comment. 
LGTM though

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -603,7 +603,9 @@
           fire-names-chan                 (u-async/call-clj-async! "get-fire-names" user-id)
           fire-cameras-chan               (u-async/call-clj-async! "get-cameras")
           user-orgs-list-chan             (u-async/call-clj-async! "get-organizations" user-id)
-          psps-orgs-list-chan             (u-async/call-clj-async! "get-psps-organizations")]
+          psps-orgs-list-chan             (u-async/call-clj-async! "get-psps-organizations")
+          fire-names                      (edn/read-string (:body (<! fire-names-chan)))]
+      (reset! !/active-fire-count         (count fire-names))
```
Oliver-BE, line 608:
```suggestion
      (reset! !/active-fire-count (count fire-names))
```

Oliver-BE:
I think the logic you added inside of `forecast_tabs.cljs` may be better suited to live inside of `near-term-forecast/get-layers!`. The last two lines of that function are currently how we decide whether or not to display a toast message depending on the layers that we get (or don't from the back-end):
```clojure
      (when-not (seq @!/param-layers)
        (toast-message! "There are no layers available for the selected parameters. Please try another combination.")))))
```
Let's just modify this to check for whether or not we're on the active fire tab. I'm thinking something like this:
```clojure
      (cond
        (and (= :active-fire @!/*forecast)
             (zero? @!/active-fire-count)
             (not (seq @!/param-layers)))
        (toast-message! "There are currently no forecasted active fires.") ;; we can always change the exact wording here later based on Shane/Chris feedback
        
        (not (seq @!/param-layers))
        (toast-message! "There are no layers available for the selected parameters. Please try another combination.")))))
```

I think we can get rid of the `!/active-fire-tab-click-count` atom and logic in order to keep consistent with the rest of the site's functionality (an error message pops up each time you navigate to a tab that doesn't have any available layers).

drewverlee:
@Oliver-BE Now the toast shows every time the layer is fetched, as per your suggestion. This seems to happen only when the forecast tab (e.g active fires) is clicked, but I didn't try to verify that it might happen at other times, as the intent in this PR is for it to happen anytime the get-layers function is called.

src/cljs/pyregence/pages/near_term_forecast.cljs
Oliver-BE: @drewverlee Can you fix the spacing on this line?

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -184,8 +184,11 @@
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
       (when-not (seq @!/param-layers)
-        (toast-message! "There are no layers available for the selected parameters. Please try another combination.")))))
-
+        (toast-message! "There are no layers available for the selected parameters. Please try another combination."))
+      (when
+          (and (= :active-fire @!/*forecast)
+               (zero? @!/active-fire-count))
+          (toast-message! "No active fires.")))))
```
Oliver-BE, line 191: Can you change this to a `cond` statement? I'm imagining something like (I haven't tested this):
```clojure
      (cond
        (and (= :active-fire @!/*forecast)
             (zero? @!/active-fire-count)
             (not (seq @!/param-layers)))
        (toast-message! "There are currently no forecasted active fires.") ;; we can always change the exact wording here later based on Shane/Chris feedback
        
        (not (seq @!/param-layers))
        (toast-message! "There are no layers available for the selected parameters. Please try another combination.")))))
```

The way you currently have it implemented it's possible for us to get both  the `"There are no layers available for the selected parameters. Please try another combination."` and `"No active fires."`, but we should only ever get one or the other.

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -615,7 +615,7 @@
           user-orgs-list-chan             (u-async/call-clj-async! "get-organizations" user-id)
           psps-orgs-list-chan             (u-async/call-clj-async! "get-psps-organizations")
           fire-names                      (edn/read-string (:body (<! fire-names-chan)))]
-      (reset! !/active-fire-count         (count fire-names))
+      (reset! !/active-fire-count (count fire-names))
```
drewverlee, line 618: @Oliver-BE I changed the spacing here, i had it the other way because you requested it above. That or misread something.

drewverlee: Looks like i misread something! I added the space in commit 7c236 and thought your suggestion was to add that same space that was already there. 

This is why I want a programmatic formatter to make these changes if possible.

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -179,11 +179,21 @@
...
                                                (:body (<! (u-async/call-clj-async! "get-layers"
                                                                                    (get-any-level-key :geoserver-key)
-                                                                                   (pr-str selected-set)))))]
+                                                                                   (pr-str selected-set)))))
+          no-param-layers?             (empty? @!/param-layers)]
       (when model-times (process-model-times! model-times))
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
       (when-not (seq @!/param-layers)
+        (toast-message! "There are no layers available for the selected parameters. Please try another combination."))
```
Oliver-BE, line 188: These lines are redundant now that the check is moved into the below `cond` statement; we can remove them.
```suggestion
```

drewverlee: @Oliver-BE redundant toast removed. Good Catch! Apologies that should have been done as part of the switch to the cond block.

Oliver-BE:
@drewverlee Pushed[ a commit](https://github.com/pyregence/pyregence/pull/896/commits/24bc3bd74194af1618d4a3ffdaa17dc733973ccc) with some small updates. The `no-param-layers?` let binding was being calculated before the `!/param-layers` atom was being reset.

drewverlee:
This makes sense to me, I never understood the role of checking the param-layers as part of deciding to trigger the active fire toast, it was [suggested early ](https://github.com/pyregence/pyregence/pull/896#issuecomment-2576155616) on and seemed to work in my test cases. 

So this looks more correct, but it's worth noting that all of this work will if I understand this [ticket ](https://github.com/pyregence/pyregence/pull/896#issuecomment-2576155616) correctly, become irrelevant as the user will never see this toast.

PR #898, opened 2025-01-02
PYR1-1026 Time slider utc toggle only affects time slider.

drewverlee:
And showing utc, outside the time slider, is done on a per options basis. e.g weather forecast start time and risk forecast start time each choose to start by showing utc. As where the rest (e.g active fires forecast start time) show local time at the start like they did before.

It's worth mentioning that as far as I can tell, the only way to toggle between utc and is the time slider toggle, so disconnecting the time slider from show-utc? might not be what we want, but it's not clear to me.

Alternatively this functionality be changed so the time slider toggle does affect the options (e.g forecast start time) but then that change (say to-local) is filtered out by a configuration setting (e.g config.cljs option would in addition to having a :default-show-utc? value would also have something like :valid-timezones #{:utc :local}. This alternative is being captured in this [PR](https://github.com/pyregence/pyregence/pull/899)

## Before commit

Here is what the weather tab looked like before this commit on load:
image

Now if you click UTC:
image

## After commit

Here is it looks like after the commit on first load:
image

After you click UTC
image

Now, as mentioned above, interperating the jira request of 1026 can go two different ways. First, what this commit does is make it so the time slider toggle only affects itself. The second notion, which this commit doesn't do, is that it keeps affecting the entire apps notion of when to show utc or not, but each spot (e.g forecast) has a way to modify that result. This could be done at the config.cljs layer.

### DIFF
--- a/src/cljs/pyregence/components/map_controls/time_slider.cljs
+++ b/src/cljs/pyregence/components/map_controls/time_slider.cljs
@@ -40,8 +40,8 @@
     [:div#time-slider {:style ($/combine $/tool $time-slider)}
      (when-not @!/mobile?
        [:div {:style ($/combine $/flex-col {:align-items "flex-start"})}
-        [radio "UTC"   @!/show-utc? true  select-time-zone! true]
-        [radio "Local" @!/show-utc? false select-time-zone! true]])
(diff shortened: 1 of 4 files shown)


drewverlee:
Given me and olivers discussion on how the feature should work, this PR shouldn't be used.

PR #899, opened 2025-01-03
PYR1-1026 Weather and Risk Tab Forecast Start time is now always UTC.

drewverlee:
This is achieved by giving each set of model parameters the option to interpret the time zone slider request to update the timezone.

For example, on the weather tab, the forecast start time option receives the request to change the timezone from :utc to :local and always returns :utc.

This change also lets each location set a different default to accommodate the feature request in 1026 that Weather and Risk Forecast Start Time always be UTC.

### DIFF
--- a/src/cljs/pyregence/config.cljs
+++ b/src/cljs/pyregence/config.cljs
@@ -694,7 +694,7 @@
                                                  :hover-text "The utility company associated with the displayed zonal statistics."
                                                  :options    {:loading {:opt-label "Loading..."}}}
                                     :model-init {:opt-label  "Forecast Start Time"
-                                                :hover-text "Start time for forecast cycle, new data comes every 6 hours."
+                                                 :hover-text "Start time for forecast cycle, new data comes every 6 hours."
                                                  :options    {:loading {:opt-label "Loading..."}}}}
                   :timezone       {:default :local
                                    :change  identity}}})
--- a/src/cljs/pyregence/config.cljs
+++ b/src/cljs/pyregence/config.cljs
@@ -206,8 +206,7 @@
                                   :model-init {:opt-label  "Model Creation Time"
                                                :hover-text "Time the data was created."
                                                :options    {:loading {:opt-label "Loading..."}}}}
-                  :timezone      {:default :local
-                                  :change  identity}}
+                  :timezone      {:change identity}}
    :fire-weather {:opt-label       "Weather"
                   :filter          "fire-weather-forecast"
                   :geoserver-key   :shasta
@@ -357,8 +356,7 @@
                                     :model-init {:opt-label  "Forecast Start Time"
                                                  :hover-text "Start time for the forecast cycle, new data comes every 6 hours."
                                                  :options    {:loading {:opt-label "Loading..."}}}}
-                  :timezone        {:default :utc
-                                    :change  (constantly :utc)}}
+                  :timezone        {:change (constantly :utc)}}
    :fire-risk    {:opt-label       "Risk"
                   :filter          "fire-risk-forecast"
                   :geoserver-key   :shasta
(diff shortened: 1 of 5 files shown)


drewverlee:
We want to change this from being per model-init and make it per "model/tab" which means changing where in the config.cljs the timezone information is kept.

drewverlee:
@Oliver-BE  You can review this now.

src/cljs/pyregence/config.cljs
```diff
@@ -203,9 +203,11 @@
...
                                                                               :filter       "cecs"
                                                                               :disabled-for #{:asp :slp :dem :cc :ch :cbh :cbd}})}
-                                  :model-init {:opt-label  "Model Creation Time"
-                                               :hover-text "Time the data was created."
-                                               :options    {:loading {:opt-label "Loading..."}}}}}
+                                  :model-init {:opt-label        "Model Creation Time"
+                                               :hover-text       "Time the data was created."
+                                               :options          {:loading {:opt-label "Loading..."}}}
+                                  :timezone   {:default :local
+                                               :change  identity}}}
```
Oliver-BE, line 210: This isn't quite the right place to put the new `:timezone` key value pair. Any values here (specifically any values inside of the map associated with the upper-level `:params` key ) will be added as a new drop down to the collapsible side panel. Here's what it looks like on my end when running your PR - there's an extra blank drop down component because of the `:timezone` addition underneath `:params`:
Screenshot from 2025-01-09 15-49-12



To understand why this happens, see the following section from `src/cljs/pyregence/components/map_controls/collapsible_panel.cljs/collapsible-panel`:

```clojure
           [collapsible-panel-section
            "layer-selection"
            [:<>
             (map (fn [[key {:keys [opt-label hover-text options sort? disabled]}]]
                    (let [sorted-options (if sort? (sort-by (comp :opt-label second) options) options)]
                      ^{:key (str (random-uuid))}
                      [:<>
                       [panel-dropdown
                        opt-label
                        hover-text
                        (get *params key)
                        sorted-options
                        (cond (ifn? disabled)     (disabled *params)
                              (boolean? disabled) disabled
                              :else               (= 1 (count sorted-options)))
                        #(select-param! % key)
                        selected-param-set]]))
                  @!/processed-params)
```

Oliver-BE: Instead, I think it would make more sense to place the `:timezone` key one level higher up so that it's at the top-level for a given tab. For the active fires tab, for example, it would look something like:

```clojure
   :active-fire  {:opt-label       "Active Fires"
                  :filter          "fire-spread-forecast"
                  :underlays       (merge common-underlays
                                          near-term-forecast-underlays
                                          {:isochrones {:opt-label        "Modeled perimeter"
                                                        :z-index          125
                                                        :filter-set       #{"isochrones"}
                                                        :dependent-inputs [:fire-name :burn-pct :fuel :model :model-init]
                                                        :disabled-for     #{:active-fires}
                                                        :geoserver-key    :trinity}})
                  :timezone   {:default :local
                               :change  identity}
                  :block-info?     true
                  :reverse-legend? true
                  :time-slider?    true
                  :hover-text      "14-day forecasts of active fires with burning areas established from satellite-based heat detection."
                  :params          {:fire-name  {:opt-label      "Fire Name"
;; ...
```

We can then read the value of `:timezone` by dereferencing the `pyregence.state/capabilities` atom. Currently, in this PR, you're dereferencing `pyregence.state/processed-params` atom to update the display text for the forecast start time timezone.

drewverlee: @Oliver-BE As suggested, I moved the timezone to the same level as the model. Thanks for catching that; it seems to have happened because my repl-browser connection went stale so I didn't see the impact of my changes (the commit was refactor so i assumed it was all good)

drewverlee: The only thing this PR probably needs is formatting changes. What formatter do we use?

My editor/formatter makes numerous small changes to the file, and it's extremely time-consuming to manually figure out how to undo them or have them follow our rule (hash map column alignment).

drewverlee:
Closing this branch in favor of this one https://github.com/pyregence/pyregence/pull/900 which is the same feature but is rebased of main. Doing this lets me keep the git history lost in the rebase.

PR #902, opened 2025-01-15
These changes are the result of applying `cljfmt fix`

drewverlee:
Here is the link to cljfmt: https://github.com/weavejester/cljfmt which I believe to be in line with the Sig Clojure Style guide.

## Purpose
To format the code base according to the sig clojure style guide and so to avoid having to ignore these changes or make them as part of other prs. Im trusting, based the Clojure style guide, that client is correct in its changes. 

## Related Issues
Closes PYR1-###

## Submission Checklist

#### Module Impacted
NA

#### Role
NA

#### Steps
Check the code changes against the Clojure style guide and play with them in your editor. 

#### Desired Outcome

Having code that follows the clojure style guide and doesn't need to be reformatted as part of other github prs. 

## Screenshots
NA


src/clj/pyregence/capabilities.clj
```diff
@@ -218,8 +218,8 @@
   [{:strs [geoserver-key workspace-name]}]
   (swap! layers
          update (keyword geoserver-key)
-                #(filterv (fn [{:keys [workspace]}]
-                            (not= workspace workspace-name)) %))
+         #(filterv (fn [{:keys [workspace]}]
+                     (not= workspace workspace-name)) %))
```
Oliver-BE, line 222: I'm not sure what the standard is, but wouldn't we also move the `(keyword geoserver-key)` portion down to its own line to be flush with the other arguments to `swap!`?

src/clj/pyregence/capabilities.clj
```diff
@@ -286,8 +286,8 @@
   (let [match-drop-names (->> (call-sql "get_user_match_names" user-id)
                               (reduce (fn [acc row]
                                         (assoc acc (:match_job_id row)
-                                                   (str (:display_name row)
-                                                        " (Match Drop)")))
+                                               (str (:display_name row)
+                                                    " (Match Drop)")))
```
Oliver-BE, line 290: Same comment here as above: should we move `(:match_job_id row)` down to its own line?

src/clj/pyregence/cameras.clj
```diff
@@ -207,16 +207,16 @@
...
   []
   (data-response
-    (if (valid-cache?)
-      @camera-cache
-      (let [new-cameras (->> (concat (get-and-conform-wildfire-cameras!)
-                                     (get-and-conform-california-cameras!))
+   (if (valid-cache?)
+     @camera-cache
+     (let [new-cameras (->> (concat (get-and-conform-wildfire-cameras!)
+                                    (get-and-conform-california-cameras!))
```
Oliver-BE, line 213: What's the reason for only having one space (instead of 2) before the sole argument to `data-response` here?

src/cljs/pyregence/pages/dashboard.cljs
```diff
@@ -141,9 +141,9 @@
      [:td ; "Delete"
       [:div {:style {:display "flex" :justify-content "center"}}
        [icon-button :trash
-                    #(handle-delete-match-drop match-job-id display-name)
-                    nil
-                    :btn-size :circle]]]]))
+        #(handle-delete-match-drop match-job-id display-name)
+        nil
+        :btn-size :circle]]]]))
```
Oliver-BE, line 146: Let's move `:trash` down to its own line. This matches what we have in various other places of the codebase. For example, in `near_term_forecast.cljs`:
```clojure
              [information-tool
               get-point-info!
               @my-box
               select-layer-by-hour!
               (get-current-layer-key :units)
               (get-current-layer-key :convert)
               (or (get-current-layer-key :no-convert) #{})
               (get-current-layer-hour)
               #(set-show-info! false)])
```

src/clj/pyregence/capabilities.clj
drewverlee: I agree. 

In general, i think cljfmt doesn't tend to move things from line to line fwiw. That would be more the realm of zprint. This allows us some flexibility and flexibility to vary case by case.

Oliver-BE: Gotcha, that makes sense!

src/clj/pyregence/cameras.clj
Oliver-BE: I see! You learn something new every day.

PR #903, opened 2025-01-17
PYR1-1031 Toggling the Layer off now removes the last toggled on Layer.

drewverlee:
## Purpose
Toggling the Layer off now removes the last toggled on Layer.

As where before toggling off the layer would internally call the server for the most recent layer in that layer category (e.g goes-16), and if that recent layer wasn't the one toggled on, the result would be a failure to remove or hide the layer.

## Further considerations

It's worth noting that further improvements can be made to the app around this fix; for one, I don't think there is a reason the client needs to call the server when toggling off. Those, and potential other changes aren't in this PR because they aren't as directly related to fixing the issue.  

## Related Issues
Closes PYR1-1031

## Submission Checklist

#### Module Impacted
Optional Layers on all forecast tabs

#### Role
ALL

#### Steps
These steps walk through re-creating the problem with toggling a layer off. The should be run first on the main branch to re-create the issue, and then on the PYR1-1031 branch to see the issue fixed.

1. Start Pyregence dev server by running `bb dev`,  and get a repl connected so you can update the running pyregence server which will be useful in later steps.
2. Click "live satellite (goes-16)" so it's toggled ON and you should see the satellite picture overlay on the map.
3. Now, to re-create the error, we need a way to show that the FE won't remove the layer its holding on to if the server doesn't respond with the same layer its currently holding. To simulate that were going to mock the response to get-layer-name. We can do this on either end, client or server, were going to do it on the server by over-riding the get-layer-name function
so eval this function to override the old one, and save the file. (I have to save the file for the repl to read the new code): `(in-ns 'pyregence.capabilities) (defn get-layer-name [geoserver-key selected-set-str] {:status 200})`
4. click the "live satellite (goes-16)" so its toggled off. On the main branch you will see that with the get-layer-name mocked the layer remains, as where on this PR its correctly removed.

In the case where the layer has just been updated, toggling that new layer on, will correctly show. All this commit does it change it so the client isnt' relying on the state of the server, which it shouldn't for hiding information it already holds.

#### Desired Outcome
Toggling the layer off now correctly removes the layer
---

## Screenshots

Screenshots from following the Steps on this branch:
With toggle on:
image
Toggle off:
image

on main branch here is the result of toggling off:
image

We see the layer stays, in production, if the layer was updated, we might see two layers at the same time!

### DIFF
--- a/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
+++ b/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
@@ -41,16 +41,21 @@
 
 (defn- toggle-underlay!
   "Toggles an underlay on the map based on the value of `show?`"
-  [filter-set dependent-inputs geoserver-key z-index show?]
+  [filter-set dependent-inputs geoserver-key z-index id show?]
   (go
     (let [cleaned-filter-set (if (seq dependent-inputs)
                                (merge-filter-set filter-set dependent-inputs) ; add in dependent-inputs to the filter-set of an underlay that isn't static
                                filter-set)
           layer-id           (<! (get-layer-name geoserver-key cleaned-filter-set identity))]
-      (when layer-id
-        (when (not (mb/layer-exists? layer-id))
-          (mb/create-wms-layer! layer-id layer-id geoserver-key false z-index))
-        (mb/set-visible-by-title! layer-id show?)))))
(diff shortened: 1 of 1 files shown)


Oliver-BE:
Really nice work on this one! I especially appreciate the detailed testing steps. :)

PR #906, opened 2025-01-22
PYR1-1030 Select the forecast after the mapbox loads.

drewverlee:
## Purpose

Selecting the forecast should be done after the mapbox loads because Mapbox asynchronously fetches styles, and likely does other work asynchronously in it's constructor, which means the app needs a way to do work after that async work finishes.

That work cannot be done before, as doing it before will cause issues like the active fire icons not loading, which is really more like a symptom of the whole mapbox program failing.

To facility doing the work explicitly after the mapbox loads, we take advantage of the Mapbox's lifecycle load event. Here is the link:

https://docs.mapbox.com/mapbox-gl-js/api/map/#map.event:load

and here is an example from that link:

```
// Initialize the map
const map = new mapboxgl.Map({});
// Set an event listener that fires
// when the map has finished loading.
map.on('load', () => {
    console.log('A load event occurred.');
});
```

To mirror this example in Pyregence, this commit passes the functionality to select the forecast to the mapbox js object called `the-map` after it's returned out of the ratom reset! in the on load spot.

I'll note that I'm not entirely sure why the mapbox object is inside a ratom in the first place, as it's already global mutable state. Is it used to somehow update facilitate rendering a component?

## Related Issues
Closes PYR1-1030

## Submission Checklist

#### Module Impacted
Mapbox

#### Role
All

#### Steps

To be clear, it's impossible to say for sure this change will fix the problem, because the problem of mapbox misbehaving and not loading things is far to general to be "fixed". Instead, our goal in these steps is just to give a good intution that doing work after the mapbox javascript object is done loading, is the right thing to do, and for the primary step in doing that, is consulting the documentation in the link above. 

Beyond that though, on the few times i saw the active fire icons fail to load, it was accompanied by this type error:

```
Uncaught TypeError: this.stylesheet is undefined
    serialize mapbox-gl-v2.10.0.js:35
    getStyle mapbox-gl-v2.10.0.js:35
    F_ core.cljs:1477
    r0 mapbox.cljs:512
    e near_term_forecast.cljs:530
    mX ioc_helpers.cljs:43
    p7 near_term_forecast.cljs:530
    nN dispatch.cljs:27
    onmessage (index):214
mapbox-gl-v2.10.0.js:35:121710
```

That type error, as you can see, stems from calling getStyleIf you try to call that function, in the commit right before this branch and right after intalizing the map, 

```clojure
(mb/init-map! "map" ...)
(.getStyle @the-map)
```

You will see the app get "stuck" likely on the "loading" screen, if you check the console, you will see the same type error. Whats likely happening is, as the pr description says, the map constructor makes an aysnc call and that call hasn't come back with any data, so the mapbox object isn't ready to answer the question "what are my styles" because it doesn't have them yet. 

This commit moves the first call i could find to get styles into the "on-load" functionality. It doesn't try to find all possible ways the map object might be called before its done loading though. That can be addressed in additional tickets.

### DIFF
--- a/src/cljs/pyregence/components/mapbox.cljs
+++ b/src/cljs/pyregence/components/mapbox.cljs
@@ -866,23 +866,27 @@
     (js/alert (str "Your browser does not support Pyregence Forecast.\n"
                    "Please use the latest version of Chrome, Safari, or Firefox.")))
   (reset! project-layers layers)
-  (-> the-map
-      (reset! (Map.
-               (clj->js (merge {:container        container-id
-                                :dragRotate       false
-                                :maxZoom          20
-                                :minZoom          3
-                                :style            (-> (c/base-map-options) c/base-map-default :source)
-                                :touchPitch       false
-                                :trackResize      true
-                                ;; For PSPS layers, we need to add basic auth to the GetTile requests
-                                :transformRequest (fn [url resource-type]
-                                                    (when (and (str/starts-with? url (:psps @!/geoserver-urls))
-                                                               (= resource-type "Tile"))
-                                                      #js {:url     url
-                                                           :headers #js {:authorization (str "Basic " (js/window.btoa (get-current-layer-geoserver-credentials)))}}))
-                                :transition       {:duration 500 :delay 0}}
-                               (when-not (:zoom opts)
-                                 {:bounds c/california-extent})
-                               opts))))
-      (.on "load" on-load-fn)))
+  (let [the-map*
+        (Map.
+         (clj->js (merge {:container        container-id
+                          :dragRotate       false
+                          :maxZoom          20
+                          :minZoom          3
+                          :style            (-> (c/base-map-options) c/base-map-default :source)
+                          :touchPitch       false
+                          :trackResize      true
+                           ;; For PSPS layers, we need to add basic auth to the GetTile requests
+                          :transformRequest (fn [url resource-type]
+                                              (when (and (str/starts-with? url (:psps @!/geoserver-urls))
+                                                         (= resource-type "Tile"))
+                                                #js {:url     url
+                                                     :headers #js {:authorization (str "Basic " (js/window.btoa (get-current-layer-geoserver-credentials)))}}))
+                          :transition       {:duration 500 :delay 0}}
+                         (when-not (:zoom opts)
+                           {:bounds c/california-extent})
+                         opts)))]
+    (.on the-map*
+         "load"
+         (fn []
+           (reset! the-map the-map*)
+           (on-load-fn)))))
--- a/src/cljs/pyregence/components/mapbox.cljs
+++ b/src/cljs/pyregence/components/mapbox.cljs
@@ -860,28 +860,29 @@
 (defn init-map!
   "Initializes the Mapbox map inside of `container` (e.g. \"map\").
    Specifies the proper project layers based on the forecast type."
-  [container-id layers get-current-layer-geoserver-credentials & [opts]]
+  [container-id layers get-current-layer-geoserver-credentials on-load-fn & [opts]]
   (set! (.-accessToken mapbox) @!/mapbox-access-token)
   (when-not (.supported mapbox)
     (js/alert (str "Your browser does not support Pyregence Forecast.\n"
                    "Please use the latest version of Chrome, Safari, or Firefox.")))
   (reset! project-layers layers)
-  (reset! the-map
-          (Map.
-           (clj->js (merge {:container   container-id
-                            :dragRotate       false
-                            :maxZoom          20
-                            :minZoom          3
-                            :style            (-> (c/base-map-options) c/base-map-default :source)
-                            :touchPitch       false
-                            :trackResize      true
-                            ;; For PSPS layers, we need to add basic auth to the GetTile requests
-                            :transformRequest (fn [url resource-type]
-                                                (when (and (str/starts-with? url (:psps @!/geoserver-urls))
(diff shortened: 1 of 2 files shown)


src/cljs/pyregence/components/mapbox.cljs
```diff
@@ -860,28 +860,29 @@
...
+                                :transformRequest (fn [url resource-type]
+                                                    (when (and (str/starts-with? url (:psps @!/geoserver-urls))
+                                                               (= resource-type "Tile"))
+                                                      #js {:url     url
+                                                           :headers #js {:authorization (str "Basic " (js/window.btoa (get-current-layer-geoserver-credentials)))}}))
+                                :transition       {:duration 500 :delay 0}}
+                               (when-not (:zoom opts)
+                                 {:bounds c/california-extent})
+                               opts))))
+      (.on "load" on-load-fn)))
```
danielhvs, line 888: Just a suggestion but you know more so feel free to ignore this suggestion:

image

danielhvs (approved):

src/cljs/pyregence/components/mapbox.cljs
drewverlee: @danielhvs Thanks for the suggestion, I want to make sure I understand it.

Before that, though, I'll reiterate that I don't understand the reason the Mapbox object is being stored inside a reagent Ratom and without that being clear to me, all my suggestions will likely be somewhat muddied.

Assuming it does need to be in a ratom then, do i understand correclty that you suggestion here is to to set (resest!) that atom after it's loaded? 

I think that would make a difference if there was a functionality that happened after the atom was reset, which could happen through `clojure.core/add-watch`. Is anything watching `the-map` ratom?

If so, then I think your suggestion is in the righ direction, that direction being that anything that relies on map being loaded some come explicitly after the map is loaded. 

I have an open question on how best to communicate that intent in the current system.

Oliver-BE: I'm not sure I understand your suggestion @danielhvs -- what functionality would this change?

danielhvs: I think a reason `the-map` is a Ratom is due to pyregence.pages.near-term-forecast/root-component:

```clojure
          (when (and @mb/the-map
                     (not-empty @!/capabilities)
                     (not-empty @!/*params))
            [control-layer user-id])
```

In all other code, `the-map` is used for side-effects or for changing styling, adding layers, making calculations, etc.
My suggestion here is just to set the (R)atom at a later time to make sure it's initialized correctly but to be honest I don't know the details so it might make not much of a difference. I think it's your or Oliver's call but from testing both ways are working fine.

Oliver-BE:
@drewverlee This PR looks good to me once you make Daniel's suggested `#(select-forecast! @!/*forecast)` change above. It's worth noting that I'm still experiencing the bug from https://sig-gis.atlassian.net/browse/PYR1-1038 on this branch.

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -632,6 +632,7 @@
       (mb/init-map! "map"
                     layers
                     get-current-layer-geoserver-credentials
+                    #(go (<! (select-forecast! @!/*forecast)))
```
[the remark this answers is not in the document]
danielhvs: I don't know much of core.async code but from my perspective and from testing:

Since:
```clojure
(defn select-forecast!
  "The function called whenever you select a new forecast/tab."
  [key]
  (go
    (do-this)
    (do-that)
    (<!
      (go
        (<! (do-bla))))))
```

So my guess is that these 2 are equivalent:
```clojure
#(go (<! (select-forecast! @!/*forecast)))
```

```clojure
#(select-forecast! @!/*forecast)
```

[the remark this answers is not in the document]
Oliver-BE: They're equivalent in this case because we don't care about the return value and we don't need to wait on `select-forecast!` to finish running to use it as the `on-load-fn` argument to `init-map!`. In general though, they might **not** have the same behavior.

If we cared about the return value or if we needed to wait for `select-forecast!` to finish, we would have to take the value from the channel that `select-forecast!` returns using the blocking take macro function `<!`. Every `go` block always returns a channel, returns that channel immediately, and starts evaluating its logic. Because all of the logic inside of `select-forecast!` is just to produce side-effects, we don't have to take anything from its result. It's enough to just return the channel from `select-forecast!` because we're throwing away that result anyways (we only care about the side-effects).

src/cljs/pyregence/components/mapbox.cljs
danielhvs: @Oliver-BE I think I've answered https://github.com/pyregence/pyregence/pull/906#discussion_r1929025904 [here](https://github.com/pyregence/pyregence/pull/906#discussion_r1929026775).

Feel free to ignore my suggestion, @drewverlee

Oliver-BE: Ah, I understand now. Thanks! Yeah, I think it can't hurt to apply your suggestion such that the ratom is set at a later date, though I can't imagine functionally much would change. @danielhvs @drewverlee can you both test out this suggestion to see if everything still works as expected in this PR?

drewverlee:
@Oliver-BE please re-review this when you have time. It has Daniel's suggestion to reset the-map atom after the map loads, though I reset and then call the select-forecast method as select-forecast might depend on the atom being set to work in some cases.

PR #907, opened 2025-01-27
PYR1-975 Small bug fix with no active fires toast message

Oliver-BE:
## Purpose
The fix to PYR1-975 wasn't quite working as expected because of an extra check in the `cond` statement altered in this PR. This is because the `no-param-layers?` check was never being hit due to the fact that the active fires icon layer exists even when it doesn't hold any data:

```clojure
cljs.user=> @pyregence.state/param-layers
[{:hour 0, :workspace "fire-detections_active-fires", :layer-group "", :forecast "fire-detections", :times ni
l, :type "active-fires", :model-init "20250127_111200", :layer "fire-detections_active-fires:active-fires_202
50127_111200", :extent ["0.0" "0.0" "-1.0" "-1.0"], :filter-set #{"20250127_111200" "fire-detections" "active
-fires"}}]
```

Simply removing that check restores the functionality that we expect. When there are no active fire forecasts (as per the `active-fire-count` atom), display the toast message that informs the user. A better solution here would be to check whether or not the `active-fires` layer in question has any data associated with it, but there's currently no easy way to do so.

## Related Issues
Closes PYR1-975

## Submission Checklist

## Testing
At the time of writing this PR (1/27/2025) there are currently no active fires. On the `main` branch, navigate to the active fires tab and see that there is no toast message that pops up (even though we would expect one). On this branch, do the same thing and see that the toast message pops up as expected.

### DIFF
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -190,16 +190,14 @@
       (when model-times (process-model-times! model-times))
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
-      (let [no-param-layers? (not (seq @!/param-layers))]
-        (cond
-          (and
-           no-param-layers?
-           (= :active-fire @!/*forecast)
-           (zero? @!/active-fire-count))
-          (toast-message! "There are currently no active fires in the system.")
+      (cond
+        (and
+         (= :active-fire @!/*forecast)
+         (zero? @!/active-fire-count))
+        (toast-message! "There are currently no active fire forecasts in the system.")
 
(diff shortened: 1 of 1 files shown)


drewverlee:
This change makes sense to me as I never fully understood the role of checking the param-layers when deciding to trigger the active fire toast; it was https://github.com/pyregence/pyregence/pull/896#issuecomment-2576155616on and seemed to work in my test cases.

So this looks more correct, but it's worth noting that all of this work will if I understand this https://github.com/pyregence/pyregence/pull/896#issuecomment-2576155616correctly, become irrelevant as the user will never see this toast.

Oliver-BE:
> it's worth noting that all of this work will if I understand this https://github.com/pyregence/pyregence/pull/896#issuecomment-2576155616 correctly, become irrelevant as the user will never see this toast.

I'm not sure I follow here. Why will the user never see the toast? And to which toast message are you referring to?

PR #909, opened 2025-01-30
PYR1-1048 show active fires forecast when there are some, otherwise show the weather tab.

drewverlee:
## Purpose

This change makes it so the browser app shows the active fire forecasts when there are some; otherwise, it sends the user to the weather forecast tab. This behavior happens on the first load.

## Related Issues
Closes PYR1-1048

## Submission Checklist

#### Module Impacted
It affects which forecast tab shows on the first load. 

#### Role
ALL

#### Steps
Run the local dev setup and hardcode the active-fire-count (line 621 in near_term_forecast.cljs) to zero. Then do a hard refresh and notice that it will take you to the weather page. Now hardcode the active-fire-count to 1 and do a hard refresh, it will go to the active fire tab as desired.

### DIFF
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -625,7 +625,19 @@
       (reset! !/user-psps-orgs-list (filter (fn [org] (some #(= (:org-unique-id org) %) @!/psps-orgs-list))
                                             @!/user-orgs-list))
       (reset! !/*forecast-type forecast-type)
-      (reset! !/*forecast (if (zero? active-fire-count) :fire-weather :active-fire))
+
+      (reset! !/*forecast
+              (cond
+                (= :long-term forecast-type)
+                (or (keyword forecast)
+                    (keyword (forecast-type @!/default-forecasts)))
+
+                ;; other wise it's near term
+                (zero? active-fire-count)
+                :fire-weather
+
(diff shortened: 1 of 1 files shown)


drewverlee:
Do we want to remove the whole notion of a forecast for near-term forecasts? It's specified in the config, then read into the client, and then set in the default-forecasts atom, which was only in the code i just removed. 

I don't have strong feelings either way.

Oliver-BE:
Nice work, @drewverlee! Let's keep around the `default-forecasts` atom because it's helpful when working locally in situations where you're working on a feature that's specific to a certain tab.

One bug I noticed: this PR breaks the `/long-term-forecast` page (http://local.pyrecast.org:8080/long-term-forecast) because the `(forecast-type @!/default-forecasts)` logic when we reset `*forecast` is removed and is instead only ever `:fire-weather` or `:active-fire` (neither of which exist on the `/long-term-forecast` page). 

Why don't we look up the value of the `*forecast-type` atom and if it's `:near-term` we can use the `(if (zero? active-fire-count) :fire-weather :active-fire))` logic and if it's `:long-term` we can use `(:long-term @!/default-forecasts)`? Alternatively we could just hard code `*forecast` to be `:fire-scenarios` when the `*forecast-type` is `:long-term` because the `/long-term-forecast` page only has one tab (which is `:fire-scenarios`).

PR #912, opened 2025-02-11
PYR1-1011 fix "dom-node" warnings

danielhvs:
## Purpose
Remove deprecated code
```clojure
(rd/dom-node this)
```

And replace with
```clojure
(.-current ref) ;; (let [ref (react/createRef)])
```

Issue: When upgrading to reagent 1.2.0 we’ve started to get deprecation notes when compiling or working with fighweel: reagent.dom/dom-node is deprecated
Fix: re-write code in such a way that fixes the warning.
Reference: re-com [issue](https://github.com/day8/re-com/issues/329)
re-com commit that fixes it: [Merge RolT PR: Removed deprecated usage of react dom node](https://github.com/day8/re-com/commit/e8a9d14)

## Related Issues
Closes PYR1-1011

## Submission Checklist

#### Module Impacted
All

#### Role
All

#### Steps
1. Configure the app to run in dev mode (call `bb dev`)
2. Test the app thoroughly. Everything should work as before.
3. Configure the app to run in prod mode (call `bb prod`)
4. Test the app thoroughly. Everything should work as before.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
         "mapbox-gl": "^3.9.3",
         "react": "^16.13.0",
         "react-dom": "^16.13.0",
-        "vega-embed": "^6.0.0"
+        "vega-embed": "^6.29.0"
       },
       "devDependencies": {
         "webpack": "^5.91.0",
@@ -140,9 +140,9 @@
       }
     },
     "node_modules/@rollup/rollup-linux-x64-gnu": {
-      "version": "4.34.3",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-x64-gnu/-/rollup-linux-x64-gnu-4.34.3.tgz",
-      "integrity": "sha512-lOyG3aF4FTKrhpzXfMmBXgeKUUXdAWmP2zSNf8HTAXPqZay6QYT26l64hVizBjq+hJx3pl0DTEyvPi9sTA6VGA==",
+      "version": "4.34.6",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-x64-gnu/-/rollup-linux-x64-gnu-4.34.6.tgz",
+      "integrity": "sha512-Sht4aFvmA4ToHd2vFzwMFaQCiYm2lDFho5rPcvPBT5pCdC+GwHG6CMch4GQfmWTQ1SwRKS0dhDYb54khSrjDWw==",
       "cpu": [
         "x64"
       ],
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "mapbox-gl": "^3.9.3",
-    "vega-embed": "^6.0.0",
+    "vega-embed": "^6.29.0",
     "react": "^16.13.0",
     "react-dom": "^16.13.0"
   },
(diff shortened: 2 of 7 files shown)


Oliver-BE:
Can you fix merge conflicts and/or rebase this on op of `main`?

danielhvs:
> Can you fix merge conflicts and/or rebase this on op of `main`?

done

Oliver-BE:
Looks good to me! I just one noticed one difference between this branch and `main` on the point info tool. The styling on this branch is slightly off -- there's an extra vertical bar of empty space on the right side of the point info tool that shouldn't be there.

On this branch:
Screenshot from 2025-02-13 11-09-37

On main (no extra empty space on the right side of the point info tool):
Screenshot from 2025-02-13 11-10-12

Oliver-BE:
@danielhvs The latest commit doesn't necessarily fix the bug above, it just turns the extra space that was added to the point info tool white. Can you take a look and see why the extra space got added in the first place? I think it looks slightly better how it was originally without this extra space on the right side of the graph.

danielhvs:
Thanks for noticing this visual difference, @Oliver-BE .
(I've just read your review about the whitespaces now BTW... I think I can take another look tomorrow, because today I spent some time digging (see below))

I've made some testing, research, looked - a lot - around and the summary is:

The vega-embed version we were using in cljsjs-vega-embed was `6.0.0` so I've tried to use that but I couldn't make it work in an easy way (without changing our compilation process). So after testing I've noticed that from version `6.26.0` on works with our regular `:require` way. I think it's because of this commit from vega-embed: https://github.com/vega/vega-embed/commit/3d1e92118635cd9e47419cd33743b1f73c6887e4
I guess before they were exporting the lib in other javscript way... Since I don't have much expertiese in JS I just think now they are exporting the lib in a way in which we can import more easily.

On top of that it seems they've changed the inner styling a little from 6.0.0 to 6.29.0 so I've just introduced a workaround to keep the background 'white'.

Here is the styling from 6.0.0: https://github.com/vega/vega-embed/blob/f4f39781f1198eec6edfe681d8645278581c863b/vega-embed.scss

But yeah, let's see how we can fix it.

danielhvs:
@Oliver-BE please re-review. This was a nice motivation for me to dive really deep in javascript/css/cljs!

I even tried to change the inner vega-embed css class in runtime but it was too overkill... (will keep this branch for a while with the testing code: https://github.com/pyregence/pyregence/compare/main...PYR1-1011-fix-node-warnings-bkp)...

But after reading vega's docs I've tried a bunch of different options to control the positions etc and ended up with this result. It's not identical but it's not better IMHO.

Oliver-BE:
Thanks for digging into this @danielhvs ! It makes sense that upgrading the Vega version changed some of the styling slightly. The latest changes still aren't quite right because the "center on selected point" button and the UI element that shows the selected point's value are now obfuscated by white:

Screenshot from 2025-02-14 12-47-45

It doesn't have to be exact, but can you do your best to match the styling from my original screenshot earlier in this PR of the `main` branch?

danielhvs:
I think now it's better, @Oliver-BE 

image


please re-re-review

danielhvs:
Or maybe this is better?

image

PR #916, opened 2025-02-21
DRAFT PYR1-1038

danielhvs:
This is just to share some findings...... Feel free to grab this code, experiment, etc ... 

The wrong behavior appears when the following events occur in the order below:

1. :reagent-render: @height: 100%
2. :component-did-mount: @height: 650px
3. "init-map!:" fn is called (will resize the map to "100%" aka 32px)
4. :reagent-render @height: 650px

If the events occur in the following order, which is more regular, then the map loads normally:

1. :reagent-render: @height: 100%
2. :component-did-mount: @height: 650px
4. :reagent-render @height: 650px
3. "init-map!:" fn is called (will resize the map to 650px correctly because the component has already be rendered with this size)

The proposal is to defer loading the map after the 2nd :reagent-render

### DIFF
--- a/src/cljs/pyregence/components/mapbox.cljs
+++ b/src/cljs/pyregence/components/mapbox.cljs
@@ -190,6 +190,7 @@
 (defn resize-map!
   "Resizes the map."
   []
+  (println "resize-map! @the-map:" @the-map)
   (when @the-map
     (js-invoke @the-map "resize")))
 
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -875,6 +875,7 @@
           (when (and @mb/the-map
                      (not-empty @!/capabilities)
                      (not-empty @!/*params))
-            [control-layer user-id])
-          [map-layer]
+            [:div
+             [control-layer user-id]
+             [map-layer]])
           [pop-up]]])})))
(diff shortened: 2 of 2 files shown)


drewverlee:
@danielhvs why are there two renders?

danielhvs:
> @danielhvs why are there two renders?

What I've written in the description is the order of events I've seen during testing, but the render fn is just one.
What happens in react is that since `:component-did-mount` changes the height (from 100% to some calculated px size, like 630px for example), react will call the `:reagent-render` fn again (2nd time).
In between these events, the map could've been initialized (drawed map in 100% height - the first value - in my case 32px).
I've noticed that the issue does not happen if the map is initialized with the "final height value" (something like 630px for me), thus after the 2nd rendering fn call.

What I did was to add `:component-did-update` to postpospone the map initialization because I've noticed this would solve the issue.
I've also removed the `init-map` fn from inside the global `initialize` fn just as a refactoring matter, but I don't think it makes much difference.

drewverlee:
@danielhvs We have two PR's for 1038, this one, and https://github.com/pyregence/pyregence/pull/919. Any ideas how to combine the ideas?

danielhvs:
> @danielhvs We have two PR's for 1038, this one, and #919. Any ideas how to combine the ideas?

I think we can close this PR and merge #919 because with #919 this "height changing issue" does even not exist anymore.

Oliver-BE:
Closed in favor of #919

PR #919, opened 2025-02-28
PYR1-1038/Improve Mapbox resizing

drewverlee:
This commit improves mapbox resizing story by calling `mapbox.resize` after mapbox loads, as opposed to on a timer which could be wrong, and by allowing the mapbox's html tag's parent height to be 100% instead of calculating it via the innerHeight.

My gut says the less Javascript involved in timing and styling the better off the browser will behave in general, and testing so far seems to show parity. However, as it's hard to test the actual problem this commit is meant to address (when the map partially loads) however, the jury is still out on how useful this change is overall.

Setting the height to 100% in the html tree necessitated some changes to the collapsible panel tree such that the "help-section" could maintain it's position at the bottom of the page.

## Testing

To reiterate, with an emphasis on testing, it's hard to test this constantly because I believe the issue was a race condition, but I'm not aware of an easy way to capture the state at various points to confirm that. What I saw was that the mapbox element would set it's height it's parent container, which in turn would be smaller then we wanted, and be set by javascript trying to figure out the height. So my intution, is that setting the height to 100% and calling "resize" after the map loads, will mean that mapbox will resize to 100% of it's parent container, which always be what we want as opposed to sometimes not being the javascript height of the window at 50ms after the component mounts.

### DIFF
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -824,46 +824,39 @@
 (defn root-component
   "Component definition for the \"Near Term\" and \"Long Term\" Forecast Pages."
   [{:keys [user-id] :as params}]
-  (let [height (r/atom "100%")]
-    (r/create-class
-     {:component-did-mount
-      (fn [_]
-        (let [update-fn (fn [& _]
-                          (-> js/window (.scrollTo 0 0))
-                          (reset! !/mobile? (> 800.0 (.-innerWidth js/window)))
-                          (reset! height  (str (- (.-innerHeight js/window)
-                                                  (-> js/document
-                                                      (.getElementById "header")
-                                                      .getBoundingClientRect
-                                                      (aget "height")))
-                                               "px"))
-                          (js/setTimeout mb/resize-map! 50))]
-          (-> js/window (.addEventListener "touchend" update-fn))
-          (-> js/window (.addEventListener "resize"   update-fn))
-          (initialize! params)
-          (update-fn)))
-      :reagent-render
-      (fn [_]
-        [:div#near-term-forecast
-         {:style ($/combine $/root {:height @height :padding 0 :position "relative" :overflow :hidden})}
-         [message-box-modal]
-         (when @!/loading? [loading-modal])
-         [message-modal]
-         [nav-bar {:capabilities     @!/capabilities
-                   :current-forecast @!/*forecast
-                   :is-admin?        (->> @!/user-orgs-list
-                                          (filter #(= "admin" (:role %)))
-                                          (count)
-                                          (< 0)) ; admin of at least one org
-                   :logged-in?       user-id
-                   :mobile?          @!/mobile?
-                   :user-orgs-list   @!/user-orgs-list
-                   :select-forecast! select-forecast!
-                   :user-id          user-id}]
-         [:div {:style {:height "100%" :position "relative" :width "100%"}}
-          (when (and @mb/the-map
-                     (not-empty @!/capabilities)
-                     (not-empty @!/*params))
-            [control-layer user-id])
-          [map-layer]
-          [pop-up]]])})))
+  (r/create-class
+   {:component-did-mount
+    (fn [_]
+      (let [update-fn (fn [& _]
+                        (-> js/window (.scrollTo 0 0))
+                        (reset! !/mobile? (> 800.0 (.-innerWidth js/window)))
+                        (js/setTimeout mb/resize-map! 50))]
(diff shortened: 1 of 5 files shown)


src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -824,46 +824,39 @@
...
-            [control-layer user-id])
-          [map-layer]
-          [pop-up]]])})))
+  (r/create-class
+   {:component-did-mount
+    (fn [_]
+      (let [update-fn (fn [& _]
+                        (-> js/window (.scrollTo 0 0))
+                        (reset! !/mobile? (> 800.0 (.-innerWidth js/window)))
+                        (js/setTimeout mb/resize-map! 50))]
```
danielhvs, line 833: Do we still need this? Jusk asking, since the map resizes on load now.

danielhvs (approved):
LGTM. I've looked around, made some testing etc.
Please check my comment regarding resizing with the 50ms timeout. I'm not sure we still need it but you might have the answer.

Video 1: comparing main and this PR

https://github.com/user-attachments/assets/4f5e8e29-d7f6-4e32-8fb4-1f7f0283bce8



Video 2: double checking the "hard resizing" in the browser using "mouse wheel"


https://github.com/user-attachments/assets/d0d5b094-7487-4aa4-960e-b65e859fc94f

src/cljs/pyregence/pages/near_term_forecast.cljs
drewverlee: @danielhvs the short answer is: i don't know because i don't understand why the timeout was there in the first place. This is kind of the problem with timeouts, its very unclear what they are trying to communicate. 

I think that resize-map! function can go away regardless, i feel like i would rather it just throw an error if the map isn't there and we can't call re-size. I don't understand the idea of returning nil and ignoring the error. Unless it somehow halts the program or something?

danielhvs: Just to make a comment, I'm not asking for changes but feel free to do so:
Agreed. I think we can drop the (when ...) and just call .resize and if it fails, something is wrong and we can see it, rather than throwing dust below the carpet.

danielhvs:
Please remember to rebase this PR on top of main and re-teste before merging.

Oliver-BE:
A couple of initial thoughts/things I noticed:

- The map now has an extra "flash" every time you load the page. This is the map manually resizing (which seemingly fixes the bug at hand). It's definitely not the most ideal behavior, but I do think it's better than occasionally getting the bug where the map doesn't load fully.
- Why is the new div on lines 307-311 of `src/cljs/pyregence/components/map_controls/collapsible_panel.cljs` needed?
- This PR introduces a bug where the top portion of the side panel is now scrollable (it shouldn't be). See before:
before
and after:
after
- This PR introduces a bug where we are unable to scroll on our other non-home pages such as http://local.pyrecast.org:8080/terms-of-use and the http://local.pyrecast.org:8080/admin. I think either lines 63-65 of `src/cljs/pyregence/components/page_layout.cljs` or the changes in `src/cljs/pyregence/styles.cljs` may be the culprit.

Oliver-BE:
There's still some styling/scrolling weirdness with Pyrecast after the latest set of commits:
Screenshot from 2025-03-12 10-28-42

drewverlee:
A couple of fixes in the latest commits seem to bring everything to where we want it to be!

According to Daniels testing, the original issue where the mapbox didn't fill the space goes away with JUST setting the mapbox canvas parent height to 100%. Originally, I thought we needed to call re-size on Mapbox load, but we don't; calling re-size was causing the extra flash, so I removed it.

Here is how it looks when it loads now:

https://github.com/user-attachments/assets/c13ae0fa-7b76-4b74-8b2f-d5c2f62e8a6b

However, putting the parent height at 100% meant fiddling with the sidebar, and that caused some other bugs, which are subsequently fixed (back to original behavior) Specifically," learn more..." is always at the bottom:

image

and it displays as before when pushed up AND doing so doesnt overflow into the header:

image

Also, the scroll is back on terms of use and admin:

image
image

Oliver-BE:
Nice work, @drewverlee! Everything is working well for me except one more small CSS bug I noticed on the Match Drop Dashboard (`/dashboard`) page. The "Refresh" button should stay at the top of the table on the `/dashboard` page when you scroll, but when I scroll it "sticks" to the top of the page:

Screenshot from 2025-03-17 11-42-25

To view the Match Drop dashboard page you'll have to be signed into an account with Match Drop access (the admin@all-orgs.com account has access).

PR #923, opened 2025-03-10
PYR1-1080  Add dev-norepl to bb to start without repl as an option

sifbiri:
## Purpose

We add an option to start the cljs compilation/figwheel server in dev mode without having to start the repl for folks who don't want repl or experiencing weird timeout errors when repl is on (like me). 

## Related Issues
Closes PYR1-1080


danielhvs:
Freedom for all. 👍

danielhvs (approved):

PR #928, opened 2025-03-19
PYR1-1075 add pyrecast tab usage GA4 events

danielhvs:
## Purpose
Start tracking forecast's clicks  (weather, risk, PSPS, active-fires and fuels)

## Related Issues
Closes PYR1-1075

## Submission Checklist

#### Module Impacted
All

#### Role
All

#### Steps
1. To really test you could create a google analytics account for yourself
2. Then configure it and get a G-* tag
3. Then deploy pyrecast in DEV (goshawk) with your G-* tag in the config.edn file (:triangulum.views/gtag-id entry)
4. Then access pyrecast in DEV (<a href="http://pyregence-dev.sig-gis.com" rel="nofollow">http://pyregence-dev.sig-gis.com</a>), click some forecasts tabs to trigger the event
5. And check the event in google analytics at Reports -&gt; Realtime Overview -&gt; "Event count by Event name" card (see Screenshots)

Personally I don't see the point of spending time with the Steps above unless you really want to because I've tested myself.

#### Desired Outcome
Events are sent to google and appear in google analytics

## Screenshots

(in my own account I've created for testing: funsimpleapps)
image

(the events appear)
image

(after clicking the event)
image

(after clicking the :forecats-type property)
image

Note: from the picture it would be 11, but now it has 6 events (last pictures). So I think google keeps tracks of the last events, not overall, at least in this report type I show in the pictures. After I refreshed the google analytics page it refreshed the count from 11 to 6.

### DIFF
--- a/config.default.edn
+++ b/config.default.edn
@@ -69,7 +69,7 @@
                                      [:meta {:name "theme-color" :content "#ffffff"}]]
  :triangulum.views/static-css-files ["/css/style.css" "/css/mapbox-gl.css"]
  :triangulum.views/static-js-files  []
- :triangulum.views/gtag-id          "G-850QRLWH9G"
+ :triangulum.views/gtag-id          "<G-SOMETHING->"
  :triangulum.views/cljs-init        pyregence.client/init
  :triangulum.views/client-keys      {:default-forecasts {:near-term :fire-weather
                                                          :long-term :fire-scenarios}


src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -550,7 +551,7 @@
                                       (assoc layer
                                              :date (u-time/get-date-from-js js-time @!/show-utc?)
                                              :time (u-time/get-time-from-js js-time @!/show-utc?)))
-                                    @!/last-clicked-info))
```
danielhvs, line 553: see f6618198ffa4840ea0254dc47b1b0ec07e2f5a6a

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -530,6 +530,7 @@
   (go
     (!/set-state-legend-list! [])
     (reset! !/last-clicked-info nil)
+    (js/gtag "event" "select-forecast" (clj->js {:forecast-type (str key)}))
```
danielhvs, line 533: We could use other names in replacement of `select-forecast` and `forecast-type`
Also we could decide on wether use `(str key)` , `(name key)` or any other mapping for the values.
From what I read and tested it seems we really need to pass a map (js obj) here otherwise it will not work.
Example that does not work: `(js/gtag "event" "select-forecast" (str key))`

sifbiri:
Looks good. I just get `gtag is not defined`, maybe it is not imported or loaded?

danielhvs:
> Looks good. I just get `gtag is not defined`, maybe it is not imported or loaded?

For that to work you need to configure the tag for triagulum to set it up:
So you need to add 
 :triangulum.views/gtag-id          "G-850QRLWH9G"
 in your config.edn
 
 Then triangulum is going to automatically add an html header: https://github.com/sig-gis/triangulum/blob/main/src/triangulum/views.clj#L105-L110
 
```clojure
     (when gtag-id
       (list [:script {:async true :src (str "https://www.googletagmanager.com/gtag/js?id=" gtag-id)}]
             [:script (str "window.dataLayer = window.dataLayer || [];"
                           "function gtag() { dataLayer.push(arguments); }"
                           "gtag('js', new Date());"
                           "gtag('config', '" gtag-id "', {'page_location': location.host + location.pathname});")]))
```

danielhvs:
> Personally I don't see the point of spending time with the Steps above unless you really want to because I've tested myself.

I've found that we can just test locally regularly and the event will also appear in google analytics normally so we don't really need to deploy to DEV to test it.

PR #932, opened 2025-04-07
PYR1-1069  Rename `dev-mode` `config.edn` key to `show-disclaimer` across Pyrecast

sifbiri:
## Purpose

Use show-disclaimer instead of dev-mode in the config and update the usage in the cljs state code. 

## Related Issues
Closes PYR1-1069


sifbiri:
> Please change also here (or remove? IDK): https://github.com/pyregence/pyregence/blob/PYR1-1069/src/clj/pyregence/handlers.clj#L45

Thanks, I have updated.

PR #934, opened 2025-04-25
PYR1-1086 Enhancing IAM and security across the site

Oliver-BE:
## Purpose
This PR enhances security across the back end by using the Ring session to perform IAM across all API routes that use the `clj-handler` handler function. The approach I went with keeps the user info on the server side inside of a Ring session instead of embedding user info inside of a JWT that the front end would send on every request, 

My implementation does this:
- Server-side session management (the session holds `user-id`, `user-email`, `match-drop-access?`, and `super-admin?` status).
- Every request uses the session to look up trusted user information.
- Authorization decisions (admin of org, super admin, etc.) are enforced server-side.

This gives us more centralized control/ease in revoking access (if/when necessary). There's a lot less overhead to deal with on the front-end and we have full control on the back-end.

## Related Issues
Closes PYR1-1086

## Submission Checklist

## Testing
There is too much here that needs testing to list given that the way in which every API function is called has changed. Complete testing will involve individually testing the functionality associated with each entry in `routing.clj` both from the REPL and via the web UI.

### DIFF
--- a/README.org
+++ b/README.org
@@ -17,7 +17,7 @@ behavior models:
 - [[https://clojure.org/guides/getting_started][Clojure CLI tools]]
 - [[https://www.postgresql.org/download][PostgreSQL (version 13)]]
 - [[https://postgis.net/][PostGIS (version 3)]]
-- [https://docs.npmjs.com/downloading-and-installing-node-js-and-npm][NPM (version 9.2.0)]]
+- [[https://docs.npmjs.com/downloading-and-installing-node-js-and-npm][NPM (version 9.2.0)]]
 
 ** Application Configuration
 
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -33,14 +33,18 @@
 
 (defn set-user-password [_ email password reset-key]
   (if-let [user (first (call-sql "set_user_password" {:log? false} email password reset-key))]
-    (data-response "" {:session {:user-id    (:user_id user)
-                                 :user-email (:user_email user)}})
+    (data-response "" {:session {:user-id            (:user_id user)
+                                 :user-email         (:user_email user)
+                                 :match-drop-access? (:match_drop_access user)
+                                 :super-admin?       (:super_admin user)}})
     (data-response "" {:status 403})))
 
 (defn verify-user-email [_ email reset-key]
   (if-let [user (first (call-sql "verify_user_email" email reset-key))]
-    (data-response "" {:session {:user-id    (:user_id user)
-                                 :user-email (:user_email user)}})
+    (data-response "" {:session {:user-id            (:user_id user)
+                                 :user-email         (:user_email user)
+                                 :match-drop-access? (:match_drop_access user)
+                                 :super-admin?       (:super_admin user)}})
     (data-response "" {:status 403})))
 
 (defn add-new-user
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -131,10 +131,9 @@
       (data-response (str "Match drop access updated to " match-drop-access?))
       (data-response "Match drop access was not able to be updated." {:status 403}))))
 
-;; TODO we can get the match drop access from the session directly
 (defn get-user-match-drop-access [session]
-  (let [user-id (:user-id session)]
-    (if (sql-primitive (call-sql "get_user_match_drop_access" user-id))`
+  (let [{:keys [user-id match-drop-access?]} session]
+    (if match-drop-access?
       (data-response (str "The user with an id of " user-id " has Match Drop access."))
       (let [response-msg (if (nil? user-id)
                            "There is no user logged in. Match Drop will remain disabled."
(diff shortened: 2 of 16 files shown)


sifbiri:
Nice work Oliver, looking good in the way it is shaping up. Some points I have: 

* In `get-current-user-email`, maybe we should also put the email in the session when the user logs in and just don’t need to have to do this call at all (including the route) - I don’t see where the user updates their email and probably if they did, they will probably have to re-login (new session when verifying)? 

* I have also noticed we are pretty much using Clojure functions for all our API routes. I know there is a SQL counterpart, but we don’t use it. Maybe we can simplify and assume all routes map to Clojure functions by default - so we drop “/clj/“ prefix in the URL routing table (we can update the cljs) and also we keep repeating the call to `clj-handler`, a higher-order function — maybe we can make it into middleware/wrapper by default? 
  * For that, maybe we can use our own `:triangulum.server/handler ` instead of the optional/default one provided by Triangulum, ` triangulum.handler/authenticated-routing-handler` (it also gives us more flexibility/control).

lambdatronic:
@sifbiri Thanks for thinking about ways to continue simplifying and securing our back-end code. I'd just like to note some things about how Pyrecast's routing table is set up currently that you'll need to preserve if you start refactoring that part of the system:

1. Every route doesn't map to a Clojure function. Note in particular the difference between Page routes and API routes. Thus, if you wanted to use a middleware to try and handle API Routes, you would need to inject something else into each request to somehow designate that it should be picked up by that middleware.

2. The optional `triangulum.handler/authenticated-routing-handler` function provides the current `routing.clj` + `handlers.clj` framework for handling back-end routes. While somewhat verbose, the `routes` map in `routing.clj` is declarative and quite straightforward to navigate by design. In particular, every route is able to declaratively define its own authentication rules. Again, using a middleware to intercept the API routes will require you to implement a new solution to this same requirement elsewhere.

3. You are correct that you can remove the `/clj/` prefix from the API routes if you continue to use the current routing table system as long as no API route names clash with Page routes. As you noted, you would need to then update `pyrecast.utils.async-utils/call-clj-async` accordingly.

4. If you want to inject additional middleware around `triangulum.handler/authenticated-routing-handler`, you can define them in Pyrecast's `handlers.clj` and then `def` them to a new var (e.g., `production-app-handler`) in the same file like so:

```clojure
(def production-app-handler (wrap-my-awesome-middleware triangulum.handler/authenticated-routing-handler))
```

You would then use `pyregence.handlers/production-app-handler` in your `config.edn`.

If that ultimately doesn't suit your needs and you think an alternative handler function is a better choice, then you can, of course, do that with the existing framework. I just wanted to make sure that you don't have to start over from scratch if that isn't necessary.

sifbiri:
> @sifbiri Thanks for thinking about ways to continue simplifying and securing our back-end code. I'd just like to note some things about how Pyrecast's routing table is set up currently that you'll need to preserve if you start refactoring that part of the system:
> 
> 1. Every route doesn't map to a Clojure function. Note in particular the difference between Page routes and API routes. Thus, if you wanted to use a middleware to try and handle API Routes, you would need to inject something else into each request to somehow designate that it should be picked up by that middleware.
> 2. The optional `triangulum.handler/authenticated-routing-handler` function provides the current `routing.clj` + `handlers.clj` framework for handling back-end routes. While somewhat verbose, the `routes` map in `routing.clj` is declarative and quite straightforward to navigate by design. In particular, every route is able to declaratively define its own authentication rules. Again, using a middleware to intercept the API routes will require you to implement a new solution to this same requirement elsewhere.
> 3. You are correct that you can remove the `/clj/` prefix from the API routes if you continue to use the current routing table system as long as no API route names clash with Page routes. As you noted, you would need to then update `pyrecast.utils.async-utils/call-clj-async` accordingly.
> 4. If you want to inject additional middleware around `triangulum.handler/authenticated-routing-handler`, you can define them in Pyrecast's `handlers.clj` and then `def` them to a new var (e.g., `production-app-handler`) in the same file like so:
> 
> ```clojure
> (def production-app-handler (wrap-my-awesome-middleware triangulum.handler/authenticated-routing-handler))
> ```
> 
> You would then use `pyregence.handlers/production-app-handler` in your `config.edn`.
> 
> If that ultimately doesn't suit your needs and you think an alternative handler function is a better choice, then you can, of course, do that with the existing framework. I just wanted to make sure that you don't have to start over from scratch if that isn't necessary.

Thank you, Gary! That is very illuminating.

lambdatronic:
@sifbiri No problem, Sif. We devised the `triangulum.handler/authenticated-routing-handler` system after studying our approaches on CEO, CAMEO, and Pyrecast and trying to take the best parts of their individual routing solutions. We are now using it on most of our Clojure web apps (CEO, CAMEO, Pyrecast, Pulse) though I believe you had created a different front-end routing solution for CoMiMo, RAMI, and SARFDAS. If you see something important that it is lacking, let us know, and perhaps we can add a patch to Triangulum to make it more feature-full. Happy hacking!

sifbiri:
@lambdatronic We are using `triangulum.handler/authenticated-routing-handler` for COMIMO and its friends too. In FCF, we also adopted it, using `cljs-http` to send JSON. So maybe leaving `triangulum.handler/authenticated-routing-handler` is a good idea.

lambdatronic:
Ah, okay! That makes sense. Glad to know that all of our web apps are unified on their back-end approach. That's quite the relief. I hope it makes it easier for everyone to jump between web apps without needing to re-learn too much about the back-end framework that they are using.

PR #939, opened 2025-05-09
PYR1-908/Hide Isochrones layer if it's dependent-type has changed

drewverlee:
## Purpose
Hide Isochrones layer if it's dependent-type has changed.

## Related Issues
Closes PYR1-908

## Submission Checklist

#### Steps

  0. Checkout git commit 253befd81e6200fc3a418042ec944816f3a6b686
  1. Go to [PyreCast.org|http://PyreCast.org] and select Active Fires from the tab options
  2. Select an active fire from the Fire Name dropdown (I suggest AZ Pim Bryce as some others are missing)
  3. Check the Modeled perimeter box from the Optional Layers panel
  4. Select a different option from the Output dropdown
  5. The Modeled perimeter that appeared in step 3 will disappear
  6. Uncheck the Modeled perimeter box
  7. Recheck the Modeled perimeter box and the layer should reappear
  
  Now, check out the latest on this branch, repeat steps 1-7, and see that in step 5, it never disappears, which is the desired behavior.

### DIFF
--- a/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
+++ b/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
@@ -133,6 +133,8 @@
                                                         underlay         (get-in @!/capabilities [@!/*forecast :underlays id])
                                                         update-underlay? (seq (set/intersection changed-keys (set (:dependent-inputs underlay))))]
                                                     (when update-underlay?
+                                                      (when (filter-set "isochrones")
+                                                        (mb/set-multiple-layers-visibility! #"isochrones" false))
                                                       (toggle-underlay! (:filter-set underlay)
                                                                         (:dependent-inputs underlay)
                                                                         (:geoserver-key underlay)
--- a/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
+++ b/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
@@ -132,8 +132,6 @@
                                                         changed-keys     (u-data/get-changed-keys old-param-map new-param-map)
                                                         underlay         (get-in @!/capabilities [@!/*forecast :underlays id])
                                                         update-underlay? (seq (set/intersection changed-keys (set (:dependent-inputs underlay))))]
-                                                    (when (filter-set "isochrones")
-                                                      (mb/set-multiple-layers-visibility! #"isochrones" false))
                                                     (when update-underlay?
                                                       (toggle-underlay! (:filter-set underlay)
                                                                         (:dependent-inputs underlay)


Oliver-BE:
This solution introduces a few bugs:
- When changing the Predicted Fire Size drop down, the new isochrones layer should display and the old one should disappear. Now, the new one displays but the old one doesn't disappear.
- When changing the Fire Name dropdown the old isochrones layer doesn't disappear. When we select a new fire we want any old isochrones layers to disappear. 

The desired behavior is:
- Changing the Predicted Fire Size drop down grabs the new isochrones layer, displays it, and removes the old one for the given fire name and forecast start time.
- Changing the Fire Name drop down shows the appropriate isochrones layer for the new fire, the predicted fire size, and  the forecast start time.
- Changing the Forecast Start Time drop down shows the appropriate isochrones layer for the forecast start time, the predicted fire size, and the forecast start time.
- Changing the Output dropdown doesn't change the previously selected isochrones layer.

Each isochrones layer represents the final predicted fire location for a given fire, percentile, and forecast start time. Changing the output type that we see on the map shouldn't change the isochrones layer. But, changing either the fire name, percentile, model (we're only using ELMFIRE for now), weather model (we're only using Hybrid for now), fuels model (we're only using LANDFIRE for now), or forecast start time should.

drewverlee:
@Oliver-BE, this is excellent feedback, thanks.

drewverlee:
The goal as stated is:

> Changing the output type that we see
> on the map shouldn't change the isochrones layer. But, changing either the fire
> name, percentile, model (we're only using ELMFIRE for now), weather model (we're
> only using Hybrid for now), fuels model (we're only using LANDFIRE for now), or
> forecast start time should.

So, to start, need a connection between the Modeled Perimeter Isochrones layer , fire-name, burn percentile, model, and fuels model,  and forecast-start-time. Luckily, that connection already exists in the configuration as the :dependent-inputs:

    {:isochrones {:opt-label        "Modeled perimeter"
                  :z-index          125
                  :filter-set       #{"isochrones"}
                  :dependent-inputs [:fire-name :burn-pct :fuel :model :model-init]
                  :disabled-for     #{:active-fires}
                  :geoserver-key    :trinity}}

These dependent types are already used to determine if we should update the overlay, for instance if the user changes the Output from Forecast fire location to CrownFire then the app (with the 908 bug) determines it should NOT update the overlay:

    (seq (set/intersection #{:output} #{:fire-name :burn-pct :fuel :model :model-init})) ;; => nil / no update!

The 908 bug is that this isn't being used to determine if the isochrones (and so Modelded perimeter) should be hidden. This is what it looks like now, where only hiding (setting visibility to false), is only dependent on
if the filter-set contains isochrones (which it only will in the above `:isochrones` map above)

    (when (filter-set "isochrones")
      (mb/set-multiple-layers-visibility! #"isochrones" false))

So, the change is to also take into account the dependent-types, which we see DONT contain `:output` hense the nil output returned from our set intersection. That would look like this:

    (when (and update-overlay? (filter-set "isochrones"))
      (mb/set-multiple-layers-visibility! #"isochrones" false))

So before, we were hiding the MP when the Output switched because the above when conditional would be true `(filter-set "isochrones") => true` during that switch BECAUSE it did it regardless of it's dependent-inputs. So once we factor that in, we get the desired behavior.

And because in the codes context we already consider updating the overlay when toggling the underlays we can nest our conditional to be more clear:

    (when update-underlay?
      (when (filter-set "isochrones")
        (mb/set-multiple-layers-visibility! #"isochrones" false))
      (toggle-underlay! ...))

This seems (crosses fingers) gives us the desired behavior Oliver

-   [Pass] Changing Output keeps Modeled perimter
-   [Pass] Changing Predicted fire size should display new isochrones layer and old one should dissaper
-   [Pass\*] When changing the Fire Name dropdown the old isochrones layer doesn't
    disappear. When we select a new fire we want any old isochrones layers to
    disappear.
-   [pass] Changing the Predicted Fire Size drop down grabs the new isochrones layer,
    displays it, and removes the old one for the given fire name and forecast
    start time.
-   [pass] Changing the Fire Name drop down shows the appropriate isochrones layer for
    the new fire, the predicted fire size, and the forecast start time.
-   [pass] Changing the Forecast Start Time drop down shows the appropriate isochrones
    layer for the forecast start time, the predicted fire size, and the forecast
    start time.
-   [pass] Changing the Output dropdown doesn't change the previously selected isochrones
    layer.

PR #952, opened 2025-07-15
PYR1-1090/add-login-analytics

drewverlee:
Closes PYR1-1090 by adding User login analytics support to the Webapp by providing a login CSV to authorized Analysts through the URL PATH /clj/get-all-users-last-login-dates.

To test this feature, first try to get the login info without proper access and notice it successfully fails, then run the
the `build-db` `migrate`, `functions` followed by the SQL `UPDATE users SET analyst = TRUE where user_uid = <your-user-id>`. After that, visit that URL and inspect the downloaded csv.


src/sql/changes/2025-7-15-add-last-login-and-analyst-to-users.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE users
+  ADD COLUMN IF NOT EXISTS last_login_date timestamptz,
+  ADD COLUMN IF NOT EXISTS analyst boolean DEFAULT FALSE;
```
Oliver-BE, line 3: Can you add these two columns to `user_tables.sql`?

deps.edn
```diff
@@ -23,7 +23,10 @@
                                        :git/sha "a3a2b218e15e9144af5e9064f686d253198f0c85"
                                        :git/tag "2024.06.05-a3a2b21"}
            sig-gis/triangulum         {:git/url "https://github.com/sig-gis/triangulum"
-                                       :git/sha "1e0e6f189fe696ad3998cf0731c2fde1b72d6254"}}
+                                       :git/sha "1e0e6f189fe696ad3998cf0731c2fde1b72d6254"}
+           org.clojure/data.csv       {:mvn/version "1.1.0"}
+           semantic-csv/semantic-csv  {:mvn/version "0.2.1-alpha1"}}
+
```
Oliver-BE, line 29: Remove extra line

src/clj/pyregence/analytics.clj
```diff
@@ -0,0 +1,17 @@
...
+
+(defn get-all-users-last-login-dates
+  [_]
+  (->> (with-open [rows (StringWriter.)]
+         (->> "get_all_users_last_login_dates"
+              call-sql
+              sc/vectorize
+              (csv/write-csv rows))
+         (str rows))
+       (assoc {:status  200 :headers {"Content-Type" "text/csv"}} :body)))
```
Oliver-BE, line 17:
```suggestion
       (assoc {:status 200 :headers {"Content-Type" "text/csv"}} :body)))
```

drewverlee:
> Do we want to introduce an analyst role as a part of this scope of work? My gut is that the super admin role covers what we want analysts to be able to do. We can always revisit this as a part of [PYR1-1109](https://sig-gis.atlassian.net/browse/PYR1-1109).

It will work either way, like I said in the longer message, I picked analyst over super-admin when I became confused about it's purpose. the analyst role is closer in nature and name to it's purpose. 

As you said, we should, in PYR1-1109 put in a bit more structure to our role system so we can more easily add, remove, and compose roles. So I suspect both super-admin and analyst will move (off the user table...) in that effort.

drewverlee:
@Oliver-BE please feel free to review this again at your convenience. you should see all the changes you caught and additional resolved the merge conflicts and as a result, had to move the logging of the login functionality. 

Thanks for your help on this!!!!

src/sql/tables/user_tables.sql
```diff
@@ -17,7 +17,9 @@ CREATE TABLE users (
     verified           boolean DEFAULT FALSE,
     verification_token text DEFAULT NULL,
     token_expiration   TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-    match_drop_access  boolean DEFAULT FALSE
+    match_drop_access  boolean DEFAULT FALSE,
+    last_login_date    TIMESTAMP WITH TIME ZONE DEFAULT NULL
```
Oliver-BE, line 21:
```suggestion
    last_login_date    TIMESTAMP WITH TIME ZONE DEFAULT NULL,
```

src/sql/changes/2025-07-15-add-last-login-and-analyst-to-users.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE users
+  ADD COLUMN IF NOT EXISTS last_login_date timestamptz,
```
Oliver-BE, line 2:
```suggestion
  ADD COLUMN IF NOT EXISTS last_login_date TIMESTAMP WITH TIME ZONE DEFAULT NULL,
```
for consistency with user_tables.sql.

drewverlee:
@Oliver-BE Thanks for the help, both suggestions approved:  The addition of the obviously missed comma, and using the more general "TIMESTAMP WITH TIMEZONE" (timestampz  is made available in postgres only it seems).

src/clj/pyregence/authentication.clj
```diff
@@ -32,6 +34,11 @@
 
 ;;; Authentication & Session Management
 
+(defn successful-login
```
Oliver-BE, line 37:
```suggestion
(defn- successful-login
```
Let's make this a private function since it's only used in `authencation.clj`.

src/sql/functions/totp_functions.sql
```diff
@@ -48,14 +48,16 @@ CREATE OR REPLACE FUNCTION get_user_with_totp(_user_id integer)
...
+    analyst boolean
  ) AS $$
 
     SELECT u.user_uid,
            u.email,
            u.match_drop_access,
            u.super_admin,
-           t.secret
+           t.secret,
+           analyst
```
Oliver-BE, line 60:
```suggestion
           u.analyst
```
for consistency.

PR #955, opened 2025-08-07
PYR1-1149: BUG: Fix `remove-workspace` function such that it processes regex strings properly

sifbiri:
## Purpose

Updated remove-workspace to handle workspace-name regexes.

## Related Issues
PYR1-1149

### DIFF
--- a/src/clj/pyregence/capabilities.clj
+++ b/src/clj/pyregence/capabilities.clj
@@ -232,11 +232,11 @@
   ;; Test exact match removal
   (let [original-layers @layers
         test-layers {:shasta [{:workspace "fire-detections_current-year-perimeters" :layer "layer1"}
-                              {:workspace "fire-detections_historical-perimeters" :layer "layer2"}
-                              {:workspace "fire-weather-forecast_gfs_20250101_00" :layer "layer3"}]}]
+                              {:workspace "fire-detections_historical-perimeters"   :layer "layer2"}
+                              {:workspace "fire-weather-forecast_gfs_20250101_00"   :layer "layer3"}]}]
     (reset! layers test-layers)
-    (remove-workspace! nil {"geoserver-key" "shasta"
-                           "workspace-name" "fire-detections_current-year-perimeters"})
+    (remove-workspace! nil {"geoserver-key"  "shasta"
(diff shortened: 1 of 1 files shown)


Oliver-BE:
Thanks, Sif! Can you provide some REPL tests in the PR description to test this new functionality and show that workspaces are removed both when a string and a regex are provided? I'm imagining something like:
```clojure
(set-all-capabilities! nil)
(remove-workspace! nil {"geoserver-key" "shasta" "workspace-name" "fire-detections_current-year-perimeters"})
; => should remove just the "fire-detections_current-year-perimeters" from the `layers` atom.
(remove-workspace! nil {"geoserver-key" "shasta" "workspace-name" "fire-weather-forecast_.*"})
; => should remove all workspaces that begin with "fire-weather-forecast_"
```

sifbiri:
I have added it as a RCT test, Oliver.

Oliver-BE:
Thanks for adding the tests, Sif! I have a few syntax suggestions to make `remove-workspace!` easier to read and to avoid re-calculating `re-pattern` for every workspace in our layers atom. Let me know what you think:

```clojure
(defn remove-workspace!
  "Given a specific geoserver-key and a specific workspace-name, removes any
   layers from that workspace from the layers atom. The workspace-name can be
   either an exact string match or a regex pattern (e.g. 'fire-weather.*')."
  [_ {:strs [geoserver-key workspace-name]}]
  {:pre [(string? geoserver-key) (string? workspace-name)]}
  (let [workspace-name-regex (try (re-pattern workspace-name)
                               (catch Exception _ nil))
        matching-workspace?  (fn [{:keys [workspace]}]
                               (or (= workspace-name workspace) ; workspace-name is a normal string and we look for an exact match
                                   (and workspace-name-regex ; workspace name is a regex string so we look for regex matches
                                        (string? workspace)
                                        (re-matches workspace-name-regex workspace))))]
    (swap! layers
           update
           (keyword geoserver-key)
           #(vec (remove matching-workspace? %))))
  (data-response (str workspace-name " removed from " site-url ".")))
```

sifbiri:
Thanks Oliver for the suggestion.

PR #1068, opened 2026-04-02
PYR1-1127/add fire risk predicted fire size combined percentile when admin

drewverlee:
As the title says, the goal is to add the fire-risk predicted fire size combined percentile when the user is an admin. So, login as a Super Admin, visit the fire risk tab, click predicted fire size then combined. This should show you the combined view on a successful matchdrop run.


danielhvs:
LGTM appeared only for the roles `(#{"super_admin" "account_manager"} user-role)`

image

danielhvs (approved):

danielhvs:
@Oliver-BE I think it is safe to deploy. I'm not authorized to merge pyregence PRs.

PR #1072, opened 2026-04-08
bugfix: add scrollbar for privacy-policy and terms-of-use pages

danielhvs:
Visit http://local.pyrecast.org:8080/terms-of-use and http://local.pyrecast.org:8080/privacy-policy pages; use dev-tools (F12 in your browser) to try new layouts; see scrollbar appears (and also footer appears!)


drewverlee:
@Oliver-BE i confirmed this works.

Oliver-BE (approved):

PR #1073, opened 2026-04-09
 PYR1-1462 bugfix match drop in pyrecast state syncing

danielhvs:
## Purpose

Fix match-drop polling silently dying on errors, leaving jobs permanently stuck "in progress".

- Wrap both polling loops in try/catch so transient exceptions (network blips, DB hiccups, bad JSON) are logged and retried instead of silently killing the future.
- Add 30s socket/connection timeouts to `poll-job!` to prevent the polling thread from hanging indefinitely on unresponsive sig3.
- Fix the timeout branch to mark the match job as failed (`md-status=1`) with a user-visible message instead of just printing to stdout and orphaning the job.
- Detect missed STARTED edges in `calculate-transitions` — when a step completes between two polls (skipping the `pending` phase), a synthetic STARTED transition is now emitted so the recorded history is always consistent.
- Extract `poll-with-retries!` to deduplicate the shared retry/timeout/error-handling loop used by both `start-polling-results!` and `poll-delete-match-drop-results-then-remove-from-db!`.
- Add unit tests for `calculate-transitions` (7 tests, 13 assertions).

## Related Issues
Closes PYR1-1462

## Submission Checklist

#### Module Impacted
Match Drop

#### Role
Admin or User with match-drop access

#### Steps
1. Trigger a match drop and tail the server log. Confirm STARTED/DONE messages arrive for every step.
2. Point `sig3-endpoint` at an unreachable host. Confirm the polling future logs errors and keeps retrying instead of dying silently.
3. Set `poll-timeout-ms` to a small value. Confirm `match_jobs.md_status` is set to 1 (failed) with a timeout message, not left at 2 (in progress).
4. Run unit tests with `clojure -M:test` and confirm 7 tests pass with 0 failures.

#### Desired Outcome
- Polling never dies silently; transient errors are logged and retried.
- Timed-out jobs are marked as failed with a user-visible message.
- Every step shows a STARTED message before DONE/FAILED, even when polling is slower than the step.

---

## Screenshots
N/A (backend-only changes)

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -60,8 +60,9 @@
            :server           {:main-opts  ["-m" "triangulum.server"]}
            :systemd          {:main-opts  ["-m" "triangulum.systemd"]}
            :test             {:extra-paths ["test"]
-                              :main-opts   ["-e" "(require,'[clojure.test,:refer,[run-tests]],'[pyregence.match-drop-test])"
-                                            "-e" "(let,[{:keys,[fail,error]},(run-tests,'pyregence.match-drop-test)],(System/exit,(if,(zero?,(+,fail,error)),0,1)))"]}
+                              :extra-deps  {io.github.cognitect-labs/test-runner {:git/tag "v0.5.1" :git/sha "dfb30dd"}}
+                              :main-opts   ["-m" "cognitect.test-runner"]
+                              :exec-fn     cognitect.test-runner.api/test}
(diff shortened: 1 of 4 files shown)


danielhvs:
timeout case (reduced timeout to test):
image

danielhvs:
I've added scrolling in the last 2 commits to improve UX

Oliver-BE (approved):

PR #1074, opened 2026-04-09
Make page layout responsive for scrollable content pages

danielhvs:
fixes up "blank bar" on top of the footer for "big screens"
NOTE this change messes with the layout of "all pages" so take a look and see if this change feels right.


danielhvs:
this breaks account-setting page layout I guess...

danielhvs:
I think we'd need a dedicated "responsive UX task for this kind of change...."

PR #1076, opened 2026-04-10
Add marketplace subscription lifecycle with Pub/Sub, entitlement API,…

sifbiri:
## Purpose

Extends the Google Cloud Marketplace integration from #1045 with subscription lifecycle via Pub/Sub, tier enforcement, SSO login, and signup fixes.

## Related Issues

PYR1-1131 PYR1-1297 PYR1-1401 PYR1-1402 PYR1-1403 PYR1-1404 PYR1-1405 PYR1-1433

## How It Works

Google POSTs a signed JWT to `/marketplace-signup`. We validate RS256 against Google's rotating X.509 certs, check issuer/audience/expiry, and pull out the customer's Google Account ID (GAIA) and orders. GAIA is an opaque per-user ID. The JWT has no email, and `sub` (procurement account ID) can be shared across users or vary per project, so GAIA is what we key on.

We try to approve the account via Google's API, then redirect existing users to `/login?marketplace=1` and new users to `/register?marketplace=1`. On first login, `complete-signup!` creates or links the org, stamps the billing fields, and retries account approval if the early attempt failed. The org starts in `pending` and stays gated until Google confirms with `ENTITLEMENT_ACTIVE` over Pub/Sub, when the actual tier gets set.

A background subscriber polls Pub/Sub for entitlement events and keeps our database in sync. We auto-approve creation and plan change requests, set the tier on activation, save the prior tier in `pre_suspend_tier` on suspension and downgrade to free until Google re-activates with the current plan, archive the org on cancellation, and clear all marketplace fields on deletion. Failed handlers get 5 retries then we ack to prevent infinite redelivery. Poll errors back off exponentially up to 5 minutes.

Returning users who click through from Google get a JWT. We find them by GAIA and log them in. If they have 2FA we redirect to verification instead. The session pulls tier from the database at that point so mid-2FA tier changes are picked up.

`route-authenticator` checks tier against the DB only when a route requires it. A subscription hierarchy (`tier3-enterprise > tier2-pro > tier1-basic-paid > tier1-free-registered > tier0-guest`) controls access the same way the role hierarchy does for admin routes. Super-admins and account managers skip tier checks.

## What Changed

Four new files under `marketplace/`: `entitlements.clj` (Procurement API), `handlers.clj` (17 event handlers), `pubsub.clj` (pull subscriber), and `util.clj` (ID normalization), replacing the old `approval.clj`. The existing `marketplace.clj` now handles signup, `complete-signup!`, and `sso-login`. Every auth path in `authentication.clj` calls `complete-signup!` so provisioning always fires. `jwt.clj` caches Google's signing keys and refreshes them safely under concurrency. The root `handlers.clj` switches tier checks from session-stored values to a live `get_org_subscription_tier` DB read, wrapped in `delay` so the read fires only when a route requires a tier check, and extends the pre-existing subscription hierarchy to include `tier0-guest`.

On the SQL side, one migration and 9 new functions cover the lifecycle events described above plus user lookup by GAIA identity. `organizations` gets marketplace account/order IDs, status, usage reporting, and pre-suspend tier columns. `users` gets `procurement_account_id` and `google_user_identity` with a partial unique index. Some functions changed from `subscription_tier` enum params to `text`.


Oliver-BE (approved):

Oliver-BE:
Really nice work, Sif!! One small thing I noticed:

In [authentication.clj:97-101](https://github.com/pyregence/pyregence/pull/1076/files#diff-authentication.clj-L97-L101) the email isn't URL-encoded in the SSO 2FA redirect:

```clojure
(str "/verify-2fa?email=" (:user_email user) "&method=totp")
```

If the email contains `+`, `&`, or `=` (e.g., `user+tag@gmail.com`), the URL breaks — `+` becomes a space on the receiving end, so `verify-2fa` looks up the wrong email and fails silently. The registration page doesn't validate email format, and the admin add-user regex explicitly allows `+`, so these emails can exist in the DB.

This is a pre-existing issue — the normal login flow in [login.cljs:37](https://github.com/pyregence/pyregence/blob/main/src/cljs/pyregence/pages/login.cljs#L37) does the same unencoded redirect. Not something to block this PR on, but worth a small follow-up fix across both paths.

Can you open up a Jira ticket for this? cc @kbeehler-sig

PR #1078, opened 2026-04-22
PYR1-1476 simplify match-drop request builder

danielhvs:
## Purpose

Refactor of `match-drop.clj`'s sig3 request builder — no behavior change to the HTTP body sent to `/api/submit-job`.

- **Drop dead locals from `params->match-drop-args`.** `:run-hours`, `:west-buffer`, `:east-buffer`, `:south-buffer`, `:north-buffer`, and `:ignition-radius` were set here but never destructured by `match-drop-args->body`, so they never reached the sig3 payload — they only survived as noise in the `match_jobs :dps-request` audit blob. sig3's `match-drop.edn` already defaults all six to identical values, so plumbing them through would be redundant.
- **Inline `params->match-drop-args` into `match-drop-args->body`.** The two-step pipeline built a flat intermediate map only to immediately destructure it again; collapsing into one function removes the indirection. The constants `"landfire"`, `"2.4.0"`, and `200` that used to round-trip through the intermediate now sit at the leaves where they're used. `submit-match-drop-job!` returns `:match-drop-inputs` as `(:arguments request)`, so `create-match-job-using-kubernetes!` still destructures `:geoserver-workspace` from it and still stores the blob as the DB audit.
- **Bind the `"md-<id>"` prefix once.** It was duplicated between `:pyrc_fire_name` and `:geoserver-workspace`; a `fire-name` let-binding keeps the two strings from drifting.

The HTTP body is byte-identical before and after. The only observable change is the shape of the `:dps-request` audit blob in `match_jobs`: it now mirrors the sig3 `:arguments` payload (nested `pyrc_*` keys) rather than the flat intermediate. Downstream code only reads `:geoserver-workspace` off that blob, which is preserved at the top level.

## Related Issues
Closes PYR1-1476

## Submission Checklist

#### Module Impacted
Match Drop

#### Role
User

#### Steps
1. Open Pyrecast against an environment that has `:sig3-endpoint` configured.
2. Initiate a Match Drop from the UI (forecast weather, any valid ignition location/time).
3. Watch the new `match_jobs` row through the DB or the UI status until all four sig3 steps (`mdrop-dps`, `mdrop-elmfire`, `mdrop-pyretechnics`, `mdrop-geosync`) report success.
4. Inspect the `match_jobs.dps_request` JSON for the new row.

#### Desired Outcome
- The Match Drop completes end-to-end exactly as on `main` (no change in runtime, cube contents, or produced layers).
- `match_jobs.dps_request` now contains the nested sig3 `:arguments` shape (`:pyrc_fire_name`, `:pyrc_simulation_span`, `:pyrc_ignition`, `:pyrc_inputs`, `:pyrc_sim_params`, top-level `:geoserver-workspace` and `:env`) rather than the previous flat `:fire-name`/`:lat`/`:lon`/... shape.

---

## Screenshots
N/A — no UI changes.

### DIFF
--- a/src/clj/pyregence/match_drop.clj
+++ b/src/clj/pyregence/match_drop.clj
@@ -292,7 +292,6 @@
      :fuel-version         "2.4.0"
      :wx-type              wx-type
      :ignition-radius      300
-     :run-hours            72
      :model-time           model-time ; e.g. Turns "2022-12-01 18:00 UTC" into "20221201_180000"
      :wx-start-time        (u/round-down-to-nearest-hour model-time)
      :fire-name            (str "md-" match-job-id)
--- a/src/clj/pyregence/match_drop.clj
+++ b/src/clj/pyregence/match_drop.clj
@@ -276,22 +276,6 @@
 ;; Create Match Job Functions
 ;;==============================================================================
 
-(defn- params->match-drop-args [match-job-id
-                                {:keys [ignition-time lat lon wx-type]}
-                                {:keys [sig3-env]}]
-  (let [model-time (u/convert-date-string ignition-time)] ; e.g. Turns "2022-12-01 18:00 UTC" into "20221201_180000"
-    {:ignition-time        ignition-time
-     :lat                  lat
-     :lon                  lon
-     :num-ensemble-members 200
-     :fuel-source          "landfire"
-     :fuel-version         "2.4.0"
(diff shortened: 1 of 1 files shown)


danielhvs:
This is just a refactor PR

Oliver-BE (approved):

PR #1080, opened 2026-04-22
PYR1-1462 fixup sig3 ignition-time display on match drop dashboard

danielhvs:
PYR1-1462 Fixup sig3 ignition-time display on Match Drop dashboard

## Purpose
The Match Drop dashboard expects `ignition-time` as a `"YYYY-MM-DD HH:MM UTC"` string and slices it with `subs` for display. For sig3-style requests, the `dps-request` only exposes `:pyrc_ignition/:pyrc_ignition_epoch_s` (a number), so the fallback produced a numeric value that broke the "Ignition Time (UTC)" column.

This PR adds `epoch-s->utc-date` in `time-utils` — the inverse of the server-side `utc-date->epoch-s` used in `match_drop.clj` — and uses it in `match-drop-item` to coerce the sig3 epoch-seconds back into the same string shape, so the column renders consistently for both request formats.

## Related Issues
Closes PYR1-1462

## Submission Checklist

#### Module Impacted
Match Drop > Dashboard

#### Role
User (with Match Drop access)

#### Steps
1. Log in as a user that has Match Drop access and at least one sig3-style match drop (i.e. a `dps-request` shaped with `:pyrc_ignition/:pyrc_ignition_epoch_s` rather than a top-level `:ignition-time`).
2. Navigate to `/dashboard`.
3. Inspect the "Ignition Time (UTC)" column for that row.
4. Also confirm a legacy match drop (one with a top-level `:ignition-time` string on `dps-request`) still displays correctly.

#### Desired Outcome
- For sig3 match drops, "Ignition Time (UTC)" shows the ignition time in `YYYY-MM-DD HH:MM` form (first 16 chars of `YYYY-MM-DD HH:MM UTC`) instead of a raw epoch-second number or `"N/A"`.
- For legacy match drops, the column is unchanged.
- No errors in the JS console (previously `subs` would have thrown on the numeric value).

---

### DIFF
--- a/src/cljs/pyregence/pages/dashboard.cljs
+++ b/src/cljs/pyregence/pages/dashboard.cljs
@@ -106,9 +106,14 @@
                                 updated-at
                                 dps-request
                                 job-log]}]
-  (let [{:keys [common-args]} (:script-args dps-request)
-        ;; WORKAROUND for sig3 requests
-        common-args           (or common-args dps-request)]
+  (let [        ;; WORKAROUND for sig3 requests
+        lon           (or (:lon dps-request)
+                          (get-in dps-request [:pyrc_simulation_span :pyrc_simspan_center_lon]))
+        lat           (or (:lat dps-request)
+                          (get-in dps-request [:pyrc_simulation_span :pyrc_simspan_center_lat]))
+        ignition-time (or (:ignition-time dps-request)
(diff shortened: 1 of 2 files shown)


danielhvs:
Ready for review, @Oliver-BE

Oliver-BE (approved):

PR #1081, opened 2026-04-23
DRAFT! dsilva-explore-tiering-match-drop-issue

danielhvs:
src/clj/pyregence/handlers.clj — New match-drop-access? helper. Derives effective access from the session-shaped map: honors the per-user :match-drop-access? flag as an override, then
  auto-grants to super admins, account managers, and accepted org members on tier2_pro+.

  src/clj/pyregence/authentication.clj
  - Added [pyregence.handlers :as handlers] require (no circular dep — handlers.clj imports nothing from pyregence.*).
  - create-session-from-user-data now stamps the session's :match-drop-access? with the derived value. Because every downstream consumer (route-authenticator at handlers.clj:90, handler bodies at match_drop.clj:565,586, capabilities at capabilities.clj:332, the frontend atom via get-user-match-drop-access) just reads that flag, they all pick up the new semantics for free — no site-by-site edits.
  - get-all-users now computes derived access per row, so the Admin tab "Match Drop?" column shows the same state the user actually experiences. Required adding subscription_tier to the SQL output.

  src/sql/functions/user_functions.sql — get_all_users returns subscription_tier alongside the other org fields it already joins.

  What this means for SRP

  Once the migration runs and the app is redeployed:
  - Any SRP user whose org is tier2_pro+ and whose membership is accepted gets Match Drop immediately on next login — no admin toggle needed.
  - Super admins and account managers get it unconditionally.
  - The per-user DB flag still works as a forced-on override for edge cases (e.g., granting a tier1 user special access).

  What's not covered (flag for review)

  - If SRP's org is still on tier1_*, the patch won't help — the PO should confirm their subscription tier is set to tier2_pro in the organizations table.
  - The admin UI toggle (update_user_match_drop_access) only writes the raw column. For tier2+ org members it'll appear to "do nothing" because derived access stays true. If admins need a per-user revoke path, that's a follow-up.
  - No new test file was added; this repo's auth tests should be extended to cover the hierarchy.
  - SQL change requires a deploy that re-runs the function definition — coordinate with whoever handles migrations.

  Verify locally

  1. Re-apply user_functions.sql against a local DB.
  2. Log in as a tier2 org member whose match_drop_access = false — the dashboard Match Drop button should appear, and /clj/get-user-match-drop-access should 200.
  3. Log in as a tier1 org member — gate should still block.
  4. Open Admin tab, search SRP — "Match Drop?" column should show green checks for accepted members once their org is on tier2+.

## Related Issues
Closes PYR1-###

## Submission Checklist
- [ ] Included Jira issue in the PR title (e.g. `PYR1-### Did something here`)
- [ ] Code passes linter rules (`clj-kondo --lint src`)
- [ ] Feature(s) work when compiled (`clojure -M:compile-cljs`)
- [ ] No new reflection warnings (`clojure -M:check-reflection`)
- [ ] For any large features/rearchitecting of the codebase, the relevant `docs` were updated (e.g. updating `docs/pyrecast-database.md` when adding a new DB table)

#### Steps

1.

#### Desired Outcome

---

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -1,5 +1,6 @@
 (ns pyregence.authentication
   (:require [pyregence.email            :as email]
+            [pyregence.handlers         :as handlers]
             [pyregence.marketplace      :as marketplace]
             [pyregence.totp             :as totp]
             [pyregence.utils            :refer [nil-on-error]]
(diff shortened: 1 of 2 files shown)


danielhvs:
I'm guessing this is tracked in jira. Let's discuss and implement later. Keeping this up though for reference.

PR #1089, opened 2026-05-28
PYR1-1527 pyrecast private geo server capabilities timeout on large layer sets

danielhvs:
## Purpose
  The PSPS GeoServer has many layers, causing the single GetCapabilities request to time out and fail to load layers on startup. This PR loads capabilities per workspace instead of all at once, and parallelizes
  the loading. Falls back to the original behavior if workspace discovery fails.

  ## Related Issues
  Closes PYR1-1527

  ## Submission Checklist

  ## Testing
  #### Module Impacted
  Misc.

  #### Role
  Admin

  #### Steps
  1. Start the application pointing to the PSPS development GeoServer
  2. Check the logs for per-workspace capability loading messages
  3. Verify all layers load successfully

  #### Desired Outcome
  All layers from the PSPS GeoServer load without timeout errors. Logs show workspace-level loading messages (e.g. `Added 142 layers from https://... (workspace=fire_history) to https://...`).

  ## Screenshots

### DIFF
--- a/src/clj/pyregence/capabilities.clj
+++ b/src/clj/pyregence/capabilities.clj
@@ -296,7 +296,9 @@
                               (let [stdout?       (= 0 (count @layers))
                                     geoserver-url (get-config :triangulum.views/client-keys :geoserver geoserver-key)
                                     new-layers    (process-layers! geoserver-url workspace-name basic-auth)
-                                    message       (str (count new-layers) " layers from " geoserver-url " added to " (get-site-url) ".")]
+                                    message       (str "Added " (count new-layers) " layers from " geoserver-url
+                                                       (when workspace-name (str " (workspace=" workspace-name ")"))
+                                                       " to " (get-site-url) ".")]
                                 (if workspace-name
                                   (do
                                     (remove-workspace! nil {"geoserver-key"  (name geoserver-key)
--- a/src/clj/pyregence/capabilities.clj
+++ b/src/clj/pyregence/capabilities.clj
@@ -1,6 +1,7 @@
 (ns pyregence.capabilities
   (:require
    [clj-http.client     :as client]
+   [clojure.data.json   :as json]
    [clojure.edn         :as edn]
    [clojure.set         :as set]
    [clojure.string      :as str]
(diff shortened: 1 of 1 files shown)


danielhvs:
@lambdatronic what do you think about us making this the default? I mean are layers that do not have workspace associated with them? 
I'd say we could just load layers per workspace and in parallell to scale.
Also we can improve robustness with retries if loading fails.

danielhvs:
Reviewers please test on `:psps "https://geoserver-utility-dev.pyrecast.org/geoserver"`

src/clj/pyregence/capabilities.clj
```diff
@@ -313,11 +316,40 @@
...
+    (if (private-layer-geoservers geoserver-key)
+      (let [geoserver-url (get-config :triangulum.views/client-keys :geoserver geoserver-key)
+            basic-auth    (str (get-psps-geoserver-admin-username) ":" (get-psps-geoserver-admin-password))]
+        (try
+          (let [workspaces (list-workspaces geoserver-url basic-auth)]
+            (log-str "Loading " (count workspaces) " workspaces from " geoserver-url)
+            (->> workspaces
+                 (pmap (fn [ws]
+                         (set-capabilities! nil {"geoserver-key"  (name geoserver-key)
+                                                 "workspace-name" ws})))
```
drewverlee, line 347: @danielhvs before the workspace-name's value was the name of the geoserver-key, which didn't require a fetch, why do we need one now?

danielhvs: Because if we don't pass it we are going to fetch ALL layers at once. And now, passing per workspace, we fetch less layers per call and don't time out.
Observe this is only for :psps which are currently timing out in PROD.

src/clj/pyregence/capabilities.clj
```diff
@@ -313,11 +316,40 @@
...
   (doseq [geoserver-key (keys (get-config :triangulum.views/client-keys :geoserver))]
-    (set-capabilities! nil {"geoserver-key" (name geoserver-key)}))
+    (if (private-layer-geoservers geoserver-key)
+      (let [geoserver-url (get-config :triangulum.views/client-keys :geoserver geoserver-key)
+            basic-auth    (str (get-psps-geoserver-admin-username) ":" (get-psps-geoserver-admin-password))]
+        (try
+          (let [workspaces (list-workspaces geoserver-url basic-auth)]
+            (log-str "Loading " (count workspaces) " workspaces from " geoserver-url)
+            (->> workspaces
+                 (pmap (fn [ws]
```
drewverlee, line 345: @danielhvs what do you think about this note on pmap as it releates to our siutation here:

> ;; Parallel application (of 'f') does NOT mean that the result collection would
;; be sorted according to calculation time. The result collection is sorted
;; in the same way as for map, i.e. it "preserves" the items' order in the 'coll'
;; (or 'colls') parameter(s) of pmap. In other words: calculation is done parallel,
;; but the result is delivered in the order the input came (in 'coll'/'colls').

;; So, e.g. if the first item of 'coll' takes 1 hour to be processed (by 'f'), and
;; the rest requires 1 sec, nothing is delivered by pmap during the 1st hour:

source: https://clojuredocs.org/clojure.core/pmap#example-5bba934be4b00ac801ed9eb3

I'm not sure it matters, as were not actually using the resulting list from pmap right? If thats the case, maybe there is another function? Or maybe we should use `dorun` instead of `doall` (a small change in my mind but sometimes things matter??)

danielhvs: Order does not matter.

drewverlee (approved):
Thanks for looking over the questions. Hope this works!

danielhvs:
> @lambdatronic what do you think about us making this the default? I mean are layers that do not have workspace associated with them? I'd say we could just load layers per workspace and in parallell to scale. Also we can improve robustness with retries if loading fails.

image

PR #1090, opened 2026-06-01
PYR1-1501 Upgrade fast-uri to v3.1.2 to fix 2 Dependabot alerts in package-lock.json

sifbiri:
## Purpose
Fixes Dependabot alert (`fast-uri < 3.1.2`, allowlist bypass via percent-encoded path/host normalization). Dev-only, transitive via `webpack -> ajv`. Pinned to `^3.1.2` via npm override.

## Related Issues
Closes PYR1-1501

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1124,9 +1124,9 @@
       "license": "MIT"
     },
     "node_modules/fast-uri": {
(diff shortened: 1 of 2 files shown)


aliceliquori (approved):
Deps look good!

PR #1093, opened 2026-06-09
PYR1-1452 update fire styles

drewverlee:
Now the fire-icons will only display as two colors, grey if the fire is 90% contained, otherwise red. Similarly, there will be a corresponding colored radius with a size proportional to the acres burned.

Tthe Jira ticket has several more pieces of functionality that won't go into this ticket. notably, it should have a total of 4 colors/images instead of just the two above. The first "yellow" or status=forecast pending, is because it's unclear, to me, atm, how to have mapbox conditionally depend on the forest ( forecasts-exists? function below). The other color, "orange," would need to be provided with information from Cloudfire (I believe).

Despite these challenges, Katy is hoping we can get this to dev.

New styles:

image

compared to prod:

image


aliceliquori (approved):
Good work!

PR #1094, opened 2026-06-10
PYR1-1545 make capabilities loading per workspace parallel and fault tolerant

danielhvs:
## Purpose
Rework GeoServer capabilities loading so every GeoServer is loaded **per workspace** (consistently, no public/private special-casing), in **parallel**, and **resilient to transient failures**.

- **Per-workspace loading for all GeoServers** — `set-all-capabilities!` lists each GeoServer's workspaces via the REST API and calls `set-capabilities!` once per workspace, avoiding timeouts on large `GetCapabilities` responses.
- **Parallelized** — replaced the sequential `doseq` over GeoServers with `pmap` (writes target distinct keys in the `layers` atom, and `swap!` is atomic, so concurrent updates are safe).
- **Retries with backoff** — added `with-retries`; the workspace listing and each per-workspace `process-layers!` call retry up to 3× with increasing delay (2s/4s/6s) on transient failures.
- **Per-GeoServer credentials** — introduced a single `:pyregence.capabilities/geoserver-credentials` config map (one entry per GeoServer), used both to list workspaces (REST) and query `GetCapabilities`. Migrated the old standalone `:pyregence.capabilities/psps` key into it; `authentication.clj` now reads psps creds from the unified map.
- **No-credentials fallback** — a GeoServer with no configured credentials can't have its workspaces listed (REST requires auth), so it falls back to a single full `GetCapabilities` call, logged loudly to stdout since that reintroduces timeout risk.
- **Refactor** — extracted the two loading branches into `load-geoserver-by-workspace!` and `load-geoserver-fully!`, leaving `set-all-capabilities!` to just dispatch on whether credentials are configured.

## Related Issues
Closes PYR1-1545

## Submission Checklist
- [x] For any large features/rearchitecting of the codebase, the relevant `docs` were updated

#### Module Impacted
Layers

#### Role
All

#### Steps
1. Add valid REST credentials for each GeoServer under `:pyregence.capabilities/geoserver-credentials` in `config.edn` (keys must match those under `:triangulum.views/client-keys :geoserver`).
2. Start the server (the `set-all-capabilities` worker runs on startup) and watch the logs.
3. Confirm each GeoServer logs `Loading N workspaces from <url>` and layers populate (e.g. map layers render, point info works).
4. Temporarily remove one GeoServer's credentials entry and restart; confirm the loud `WARNING: No credentials configured for GeoServer …` line and that it falls back to a full `GetCapabilities`.

#### Desired Outcome
All GeoServers load their layers via per-workspace calls; transient REST/HTTP failures are retried rather than dropping layers; a GeoServer missing credentials still loads (with a visible warning) instead of silently loading nothing.

## Screenshots
N/A — server-side change, no UI.

### DIFF
--- a/config.default.edn
+++ b/config.default.edn
@@ -122,7 +122,7 @@
  :pyregence.capabilities/geoserver-credentials      {:trinity     {:username "<username>" :password "<password>"}
                                                      :shasta      {:username "<username>" :password "<password>"}
                                                      :match-drop  {:username "<username>" :password "<password>"}
-                                                     :psps        {:username "admin"      :password "password"}
+                                                     :psps        {:username "<username>" :password "<password>"}
                                                      :pyreclimate {:username "<username>" :password "<password>"}}
  :pyregence.weather-stations/get-observation-stations-every-n-minutes 1440 ;; once a day
  :pyregence.email/verification-token-expiry-minutes 15
--- a/config.default.edn
+++ b/config.default.edn
@@ -117,6 +117,14 @@
  :pyregence.cameras/alert-west-api-key              "<api-key>"
  :pyregence.capabilities/psps                       {:geoserver-admin-username "admin"
                                                      :geoserver-admin-password "password"}
+ ;; Credentials used to list workspaces (REST API) and query GetCapabilities per
+ ;; GeoServer. Omit an entry, or leave username/password blank, to query that
+ ;; GeoServer anonymously. Keys must match the keys under :geoserver above.
+ :pyregence.capabilities/geoserver-credentials      {:trinity     {:username "<username>" :password "<password>"}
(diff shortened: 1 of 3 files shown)


danielhvs:
after merging this we need to merge https://github.com/PyreCast/pyrecast-gcp-deployment/pull/20 (although it really would not "break" the app)

drewverlee (approved):
I read through this, and it looks good @danielhvs. I haven't had a chance to test it (by running it locally), which I'll try to get to in the next 2 hours, but if you're highly confident it works, I don't see why we can't get it to dev right now. Otherwise, ping me if you would like a local test of this!

danielhvs:
Thank you, @drewverlee ! 
I've added some tweaks. 
Merging.

PR #1095, opened 2026-06-17
PYR1-1529 add elmfire wui as a separate model for processing active fires

danielhvs:
## Purpose
Adds **WUI active fires** as a separate selection within the Active Fires tab. These fires are produced by the ELMFIRE WUI model and are served from the private `:psps` GeoServer (rather than `:trinity` like regular active fires). Access is restricted to members of the **Pyregence Consortium** organization and gated behind a new `:wui` feature flag.

Key changes:
- **New `:wui` feature flag** (`config.default.edn`) — gates WUI layer parsing in capabilities and the WUI Fire select in the UI.
- **WUI Fire select** — a new `:wui-fire-name` param is added to the Active Fires tab, populated from the `:psps` GeoServer and keyed off `:wui-fire-name` (parallel to `:fire-name`/`:match-drop-name`). The `:params` map was switched to an `array-map` to preserve render order now that it has 9+ entries.
- **Authenticated tiles** — WUI fire tiles live on the private `:psps` GeoServer, so they are served using the PSPS GeoServer admin credentials. A new `get-psps-geoserver-credentials` endpoint returns those credentials only to `pyregence-consortium` members (and `super_admin`s); the value is stored in `!/psps-geoserver-credentials` and used by `get-current-layer-key` when a WUI fire is selected.
- **Modeled perimeter underlay** — the isochrones "Modeled perimeter" underlay is resolved against `:psps` and re-keyed from `:fire-name` to `:wui-fire-name` when a WUI fire is selected (`effective-underlay` in `collapsible_panel.cljs`).
- **Capabilities** — `process-capabilities` filters WUI layers (`wui-` prefix) behind the flag, and `get-fire-names` now returns a `:wui-active-fires` group.
- **DB migration** — assigns the listed Consortium users to the `pyregence-consortium` org as regular members.

## Related Issues
Closes PYR1-1529

## Submission Checklist

#### Module Impacted
Side Panel > Active Fires tab; Layers (Modeled perimeter underlay); Match Drop (fire-name refresh)

#### Role
User who is a member of the **Pyregence Consortium** org and/or SuperAdmins

#### Steps
1. Ensure the `:wui` feature flag is enabled in config.
2. Log in as a **Pyregence Consortium** member.
3. Open the **Active Fires** tab and confirm a **WUI Fire** dropdown appears with WUI active fires listed.
4. Select a WUI fire and confirm the layer tiles render (served authenticated from the `:psps` GeoServer).
5. Toggle the **Modeled perimeter** underlay and confirm it renders for the selected WUI fire.
6. Switch between a regular active fire / match drop / WUI fire and confirm selecting one resets the others.
7. Log in as a user **not** in the Consortium and confirm the WUI Fire select is hidden.

#### Desired Outcome
Consortium members can view and interact with WUI active fires (including the modeled perimeter underlay) with authenticated tiles; non-members never see the WUI Fire select.

## Screenshots
wui


sifbiri:
Nice work Daniel.

`get-psps-geoserver-credentials` (`authentication.clj:897`) sends the GeoServer admin login to the browser to load WUI layers. Might be safer to give WUI its own non-admin login instead.

The migration overwrites `user_role` + `organization_rid` for the listed users with no checks, so it could demote a `super_admin` or move someone out of their org. Worth checking.

danielhvs:
Thanks, @sifbiri 
1. Migration removed after discussion.
2. Removing the admin pwd for the client by applying the approach: we're treat the pyregence consortium as an organization and add geofence rules for it. (new changes added in the GCP repo [PR](https://github.com/PyreCast/pyrecast-gcp-deployment/pull/23))

danielhvs:
worked out: 
image

src/cljs/pyregence/components/map_controls/match_drop_tool.cljs
```diff
@@ -120,7 +120,13 @@
         (swap! !/capabilities update-in [:active-fire :params :match-drop-name :options] merge (:match-drops fire-names))
         (swap! !/capabilities assoc-in [:active-fire :params :match-drop-name :hidden?] false)
         (swap! !/processed-params update-in [:match-drop-name :options] merge (:match-drops fire-names))
-        (swap! !/processed-params assoc-in [:match-drop-name :hidden?] false)))))
+        (swap! !/processed-params assoc-in [:match-drop-name :hidden?] false))
+      ;; WUI active fires (pyregence-consortium members only, gated behind the :wui feature flag)
+      (when (c/show-wui-fires? @!/user-orgs-list (:wui-active-fires fire-names))
+        (swap! !/capabilities update-in [:active-fire :params :wui-fire-name :options] merge (:wui-active-fires fire-names))
```
drewverlee: It would be better if we had one swap per atom here, not just for your changes, but for the previous ones to. Otherwise we could hit more race conditions.

src/cljs/pyregence/config.cljs
```diff
@@ -14,6 +14,33 @@
   [feature-name]
   (get @!/feature-flags feature-name))
 
+(def wui-org-unique-id
```
drewverlee: If there is any way for this to be a namespaced keyword, that would be far better than a string because clojure can help keep namespaced keywords unique and supports using them as keys.

src/cljs/pyregence/config.cljs
```diff
@@ -14,6 +14,33 @@
...
+  "The `org-unique-id` of the organization that owns WUI active fires. Only members of
+   this org may view the WUI Fire select."
+  "pyregence-consortium")
+
+(defn user-in-wui-org?
```
drewverlee: Not a big deal, but consider having a wui namespace, im not sure why these functions are config specific.

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -365,13 +365,18 @@
...
-                          :fuels        :only-underlays
-                          :fire-weather [:fire-weather :model]
-                          :fire-risk    [:fire-risk :pattern]
-                          :active-fire  :only-underlays
-                          :psps-zonal   [:psps-zonal :utility]
-                          nil)]
+    (if (c/wui-fire-selected?)
+      ;; WUI active fires are private to the pyregence-consortium org, which is just a
+      ;; regular PSPS org: authenticate with its own GeoServer credentials.
+      (some #(when (= c/wui-org-unique-id (:org-unique-id %)) (:geoserver-credentials %))
```
drewverlee: It might be clearer to turn the user-psps-org-list into a mapping from key->value, then use that to look up the value needed if it exists, rather than search with some.

The search with some is a bit more flexable and likely to break overtime for some reason.

drewverlee (approved):
looks good beyond some minor organization changes i personally think would help, at the very least, read over my comments and think about those kinds of changes for next time, but feel free to merge regardless.

Note i didn't do any kind of testing or much verification of this ticket against the jira requirements.

src/cljs/pyregence/config.cljs
danielhvs: Good idea in principle, but I don't think it works here: this value is compared with `=` against `:org-unique-id`, which comes from the DB as a string (`authentication.clj` builds orgs with `:org-unique-id org_unique_id`, read client-side from EDN). A namespaced keyword would never `=` that string, so I've kept it a string. (It does get moved into the new `pyregence.wui` namespace per your other comment.)

drewverlee: I forgot that keys and entity IDs are different in a very subtle way from keys: keys identify a category, an ID identifies a specific thing in that category.

PR #1099, opened 2026-06-17
PYR1-604/add-nfdrs-constant-and-variable-model-and-weather-params-to-weather-tab

drewverlee:
Adds NFDRS constant and variable model and their exclusive weather parameters to the weather tab if the user is an Organization Admin or Organization Member with a tier1_basic_paid subscription or above, or a Super Admin. 

This is dependent on a deployment [PR here ](https://github.com/PyreCast/pyrecast-gcp-deployment/pull/24 )to sync the layers.

### DIFF
--- a/.github/workflows/clj-kondo.yml
+++ b/.github/workflows/clj-kondo.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
-    - uses: DeLaGuardo/clojure-lint-action@master
+    - uses: actions/checkout@v4
+    - uses: DeLaGuardo/setup-clj-kondo@master
       with:
-        clj-kondo-args: --lint src --fail-level error
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        version: '2026.05.25'
+    - run: clj-kondo --lint src --fail-level error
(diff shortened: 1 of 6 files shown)


aliceliquori:
@danielhvs - We have been in a little bit of a pickle and think we are nearing the finish line, but have one conspicuous bug. In the weather tab, there are two issues.

If you can see NFDRS models, then one of the NFDRS model layers is selected by default. If we add the suspect layers to the appropriate :disabled-for form, we run into a chicken-and-egg problem where a user can never select NFDRS models or the layers they have. I am not sure of a good way to do this kind of stateful rendering where two visible items depend on each other's state without a more powerful state framework like re-frame or something. Can't select param without selecting model, but model is disabled for default params. I think this is a first of a kind issue?

I managed to slowly figure out most of the other parts of this ticket with Drew, but Id like your input here on rendering it. Check the changes in near-term-forecast.

I almost feel like we need a new parent dropdown to select which weather model we want, and then show the respective layers & etc underneath, instead of co-mingling everything the way we are.

danielhvs:
During our meeting today @drewverlee and @kbeehler-sig talked about the idea to keep a new select apart for NFDRS. It'd keep it simpler to maintain

drewverlee:
closed in favor of https://github.com/pyregence/pyregence/pull/1106

PR #1102, opened 2026-06-22
Dsilva pyr1 604 nfdrs

danielhvs:
just used AI to tackle some issue that was "straightforward"


danielhvs:
@aliceliquori @drewverlee maybe this would fix some of the issues you were having? IDK. 
Would be nice if we really have the layers in the gcp bucket to really test end to end IMHO.

danielhvs:
NOTE: this PR is to introduce code into
[PYR1-604-NFDRS](https://github.com/pyregence/pyregence/tree/PYR1-604-NFDRS)

drewverlee:
@danielhvs this doesn't fix the main issue, but it is a good idea to share the nfdrs set in a variable.

PR #1103, opened 2026-06-23
PYR1-1550 PYR1-1518 Stop user email enumeration

sifbiri:
## Purpose

The app revealed whether an email is registered. Two fixes:

- Forgot-password now returns one generic response for any email.
- Removed the `/clj/user-email-taken` endpoint and the registration pre-check that used it (the DB still enforces unique emails).

## Related Issues
Closes PYR1-1550 PYR1-1518

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -734,14 +734,6 @@
     ;; TODO what should this failure message be.
     (data-response "" {:status 400})))
 
-(defn user-email-taken
-  ([_ email]
-   (user-email-taken nil email -1))
-  ([_ email user-id-to-ignore]
-   (if (sql-primitive (call-sql "user_email_taken" email user-id-to-ignore))
-     (data-response "User email is taken")
-     (data-response "User email is not taken" {:status 403}))))
-
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Access Control
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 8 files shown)


danielhvs:
## Review — Stop user email enumeration

Clean, well-scoped security fix. The forgot-password path now returns an identical `200` + generic body regardless of account existence and normalizes timing by sending off-thread via `future`. Verified: after this PR there are no remaining references to the `user-email-taken` endpoint or the removed `pending?` atom in `login.cljs`; `users.email` is `NOT NULL UNIQUE` (the DB does enforce uniqueness as claimed); and `users.name` is `NOT NULL` with registration storing `""` (truthy), so the `when user-name` gate tracks account existence without dropping reset emails for real users.

Nothing rises to a correctness bug. A few low-severity cleanup/fragility notes:

- **`src/sql/functions/user_functions.sql:75` — `user_email_taken` SQL function is now dead code.** Its only caller was `authentication/user-email-taken`, removed in this PR, and nothing else references the endpoint. Consider dropping it (with a `DROP FUNCTION` migration) here or in a follow-up.

- **`src/clj/pyregence/email.clj` (rct test block) — test relies on un-redefed `get-config`.** The `^:rct/test` comment redefs `call-sql`, `get-email-format`, `send-mail`, and `dispatch-off-thread`, but the `send!` closure still calls `(get-config :triangulum.email/base-url)` and (since `security?` is true) `get-security-email-config` → `(get-config :mail)`. If rct runs without a loaded config, these return nil or throw, making the test environment-dependent. Redef `get-config` or stub the two lookups to make the test self-contained.

- **`src/cljs/pyregence/pages/login.cljs` `request-password!` duplicates the generic message** that also lives in `email.clj` as `generic-reset-response`. Identical today but can drift independently. Minor (the client shows its own copy and ignores the server body) — a shared constant or comment would keep them in lockstep.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

danielhvs:
@sifbiri please rebase off main or merge from main so that we can run :rct here in a CI/CD manner

sifbiri:
Thank you @danielhvs. I have updated it.

danielhvs (approved):

danielhvs (approved):

PR #1105, opened 2026-06-24
PYR1-1512 Enforce non-predictable identifiers in PyreCast

danielhvs:
## Summary

Fixes the IDOR (Insecure Direct Object Reference) exposure behind PYR1-1512 by enforcing **per-resource authorization** on the org/user admin and match-drop endpoints. Following the [OWASP IDOR Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html), the fix is an access-control check scoped to the acting user (derived from the session), not obfuscation of identifiers.

The endpoints were **role-gated but not resource-scoped**: the route layer verified a caller was *an* `organization_admin`, but the handlers then acted on whatever org/user id the client supplied. So an admin of one organization could edit another org, or change/remove users in it, by passing a foreign id. Match-drop delete/status acted on any job id regardless of owner.

## Changes

- **New SQL authorization predicates** (`organization_functions.sql`):
  - `can_admin_org(requesting_user, org_id)` — `super_admin`/`account_manager` may administer any org; an `organization_admin` only their own.
  - `can_admin_user(requesting_user, target_user)` — delegates to `can_admin_org` via the target user's org.
- **Gated the org/user admin handlers** (`authentication.clj`) on those predicates, using the acting user from the session; cross-org/cross-user requests now return `403`:
  - `update-org-info`, `add-org-user` (org-scoped)
  - `remove-org-user`, `update-org-user-role`, `update-user-org-membership-status`, `update-user-name` (user-scoped)
  - `super_admin`/`account_manager` remain unrestricted.
- **Scoped match-drop to the owner** (`match_drop.clj`): `delete-match-drop!` and `get-md-status` verify the job belongs to the session user, returning a uniform `403` that does not reveal whether the job exists.
- **Tests**: added `^:rct/test` blocks asserting the predicates allow same-org and deny cross-org access, that each guarded handler returns `403` across org boundaries, and that match-drop access to a non-owned job is denied.

No schema or primary-key changes: existing integer ids are kept; the fix is purely authorization. The predicates are `CREATE OR REPLACE` functions loaded by `bb functions` on deploy.

## Test plan

- [x] `bb test` — 22 tests, 129 assertions, 0 failures.
- [ ] As an `organization_admin`, confirm actions on your own org/users still succeed and actions targeting another org/user return `403`.
- [ ] Confirm `account_manager`/`super_admin` retain cross-org access.
- [ ] Confirm a user cannot view (`get-md-status`) or delete another user's match drop.

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -688,10 +688,12 @@
 
 (defn update-user-name
   "Allows an organization admin to update the name of a user by their email."
-  [_ email new-name]
+  [{:keys [user-id]} email new-name]
   (if-let [user-id-to-update (sql-primitive (call-sql "get_user_id_by_email" email))]
-    (do (call-sql "update_user_name" user-id-to-update new-name)
-        (data-response (str "User's name successfully updated to " new-name)))
+    (if (sql-primitive (call-sql "can_admin_user" user-id user-id-to-update))
+      (do (call-sql "update_user_name" user-id-to-update new-name)
+          (data-response (str "User's name successfully updated to " new-name)))
+      (data-response "You are not authorized to manage this user." {:status 403}))
     (data-response (str "There is no user with the email " email)
                    {:status 403})))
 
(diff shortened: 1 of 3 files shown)


src/clj/pyregence/authentication.clj
```diff
@@ -32,6 +33,51 @@
...
+   by the rich-comment test below."
+  #{:match-drop-access?
+    :user-email
+    :user-name
+    :user-role
+    :org-membership-status
+    :subscription-tier
+    :marketplace-status})
+
+(def internal-session-keys
```
drewverlee: There is no reason to set up guard rails here, we just don't pass the PostgreSQL IDs to the client, and instead pass business-level unique identifiers like emails and unique-org-names that multiple systems can agree upon.

src/clj/pyregence/handlers.clj
```diff
@@ -1,18 +1,29 @@
...
+
+(defn render-page
+  "Wraps `triangulum.views/render-page`, stripping `internal-session-keys`
+   from the session before it is embedded in the page source. The persisted
+   session cookie is unaffected (the render response sets no `:session`), so
+   server-side handlers still see these keys."
+  [uri]
+  (let [page (triangulum-views/render-page uri)]
+    (fn [request]
+      (page (apply update request :session dissoc internal-session-keys)))))
```
drewverlee: Having middleware remove keys that developers think are going to be available through the session is going to lead to confusion and bugs.

drewverlee (changes requested):
In short, we should use uesr-email instead of user-id and org-unique-name instead of organization-id.

danielhvs:
@drewverlee do we want to drop the columns from the database or just use the replacements in the application (clj and cljs)? I think we must keep the pks in the db and only change the app. 

replacements:
user-id -> user-email
organization-id -> org-unique-name

danielhvs:
@drewverlee I've started the work but seems really big if we want to change it all.
Let me know if you think I'm on the right direction. I'll pivot to work on other tasks I have to tackle.

danielhvs:
DRAFT

danielhvs:
Will open another PR later.

PR #1106, opened 2026-06-24
PYR1-604/nfrds-drew

drewverlee:
Adds NFRDS models and their weather params if the user has super-admin or their org has the right subscription role. The NFSRDS weather params will only show when the NFRDS model is selected to work a circle dependency issue that exists in other model to wp relationships, this means in order to get to other WP once the model is selected you have to change your model first.

This needs to go in with this [PR](https://github.com/PyreCast/pyrecast-gcp-deployment/pull/24)


aliceliquori (approved):
Beauty and perfection through teamwork!

PR #1107, opened 2026-06-24
PYR1-1552/remove-active-fire-background-when-not-on-active-fire-tab

drewverlee:
Just like the title says, this removes the fire-active-background (the red circles behind the active fire icons) when the user isn't  on the active fire tab.

To see this bug, visit prod on the active fire tab and then move to another tab and notice the background circle is still there...which we don't want. To see the fix, spin this up locally and do the same thing and notice its not there anymore.


kbeehler-sig (approved):

drewverlee:
That layer string should be variable and synced to where it's registered. 

I'm not sure that's the best place to remove the layer but I'm but sure what else it would be.

aliceliquori:
I'm gonna go ahead and merge this because we talked about it in our teamwork on other aspects today - good stuff, brother

PR #1108, opened 2026-06-25
only show active fire background circle when 'All Active Fires' 

drewverlee:
only show active fire background circle when 'All Active Fires'

The location and timing of this remove-layer function seems... questionable.

Broadly speaking the question is if the layer showing is connected to a UX
change: user clicks new tab and user clicks new fire, or a layer change, e.g
the user has selected an active fire layer. Right now, it seems the later, so
the current solution of binding it to the layer is better. As in the icon and the circle layer should travel together.

But then it's a
question of when to check and its very unclear "when" this get-layers function
is called, but in testing it seems ok.

But this works, and thats better then nothing.

To test it, load the active fires tab, notice the circle is removed whenever Active Fires > Fire Name isn't "All Active Fires"

### DIFF
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -242,7 +242,7 @@
       (when model-times (process-model-times! model-times))
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
-      (when-not (= (:type current-layer) "active-fires") (mb/remove-layer! "fire-active-background"))
+      (when-not (= (:type (current-layer)) "active-fires") (mb/remove-layer! "fire-active-background"))
       (cond
         (and
          (= :active-fire @!/*forecast)
--- a/src/cljs/pyregence/pages/near_term_forecast.cljs
+++ b/src/cljs/pyregence/pages/near_term_forecast.cljs
@@ -242,9 +242,7 @@
       (when model-times (process-model-times! model-times))
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
-      (let [{:keys [type forecast]} (current-layer)]
-        (when (or (not= type "active-fires") (= forecast "fire-spread-forecast"))
-          (mb/remove-layer! "fire-active-background")))
+      (when-not (= (:type current-layer) "active-fires") (mb/remove-layer! "fire-active-background"))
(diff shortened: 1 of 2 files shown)


drewverlee:
@danielhvs @sifbiri give this a look over please and thanks!!!

drewverlee:
TODO the fire-active-background layer name should be in a var if were going to re-use it on add and remove. someone, probably me, do that...

src/cljs/pyregence/pages/near_term_forecast.cljs
```diff
@@ -242,6 +242,7 @@
       (when model-times (process-model-times! model-times))
       (reset! !/param-layers layers)
       (swap! !/*layer-idx #(max 0 (min % (- (count (or (:times (first @!/param-layers)) @!/param-layers)) 1))))
+      (when-not (= (:type current-layer) "active-fires") (mb/remove-layer! "fire-active-background"))
```
danielhvs:
```suggestion
      (when-not (= (:type (current-layer)) "active-fires") (mb/remove-layer! "fire-active-background"))
```

danielhvs (changes requested):
Requesting changes on one correctness bug, plus a small lint cleanup.

### 1. Bug: `current-layer` is used as a value, not called

In `near_term_forecast.cljs`:

```clojure
(when-not (= (:type current-layer) "active-fires") (mb/remove-layer! "fire-active-background"))
```

`current-layer` is a zero-arg function (`defn- current-layer []`), not a layer map. Used bare, `(:type current-layer)` looks up `:type` on the **function object**, which always returns `nil`. So the guard is permanently `(when-not (= nil "active-fires") …)` → always true, and `fire-active-background` gets removed on **every** `get-layers!` call — including when the current layer really is "active-fires".

It only looks correct because `change-type!` re-adds the highlight right afterwards whenever `style-fn` is truthy, so the guard is effectively dead code today and will misbehave the moment that re-add path changes.

Fix — call the function (as every other use in this file does, e.g. `(:layer (current-layer) "")`):

```clojure
(when-not (= (:type (current-layer)) "active-fires") (mb/remove-layer! "fire-active-background"))
```

(See the inline comment already on this line.)

### 2. Drop the now-unused require (lint)

In `forecast_tabs.cljs`, after removing the `mb/remove-layer!` call from the `on-click` handler, `mb` is no longer referenced in the file. Remove the require to clear the lint warning:

```clojure
[pyregence.components.mapbox :as mb]   ;; delete
```

---
*I reviewed and analyzed this PR myself, and checked and tested these findings; Claude Code assisted with drafting.*

drewverlee:
@danielhvs good catch on the function call, that was a regression i made right at the end.

danielhvs (approved):

PR #1110, opened 2026-06-26
PYR1-1554 The Time Slider's time toggle and speed control now shows on mobile.

drewverlee:
As the title says, the time slider's time toggle (utc and local) and speed control (back forward and speed inc) now shows on mobile. Note this takes up mores space on horizontal mobile view, but it makes
them fit on the vertical mobile view. It doesn't change the desktop view.

To test this, do exactly that, try the different views in the browser, if you can go the extra mile and try our mobile emulator.

An easy improvement would be to be to use the space better on the horizontal view. Thanks and good luck reviewer.


danielhvs:
current:
image

with PR changes:
image

danielhvs (approved):

PR #1114, opened 2026-06-29
PYR-1555/Send nfdrs weather param query to server to get layers

drewverlee:
Now the weather param query is in request to the server. Select WP > NFDRS constant see weather parameter param in the get-layer request.

### DIFF
--- a/src/cljs/pyregence/config.cljs
+++ b/src/cljs/pyregence/config.cljs
@@ -9,10 +9,7 @@
 ;; NFDRS Weather Params
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-;; TODO consider moving this state because its only here because
-;; when you select the nfdrs weather  param it seems to use the key (e.g erc) to look up
-;; up the :filter in the config and not from the html dropdown.
-(def  nfdrs-weather-params
+(def nfdrs-weather-params
   {:erc    {:opt-label "Energy Release Component (ERC, Btu/sq ft)"
             :filter "erc"
             :units "(ERC, Btu/sq ft)"}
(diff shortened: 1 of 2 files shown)


aliceliquori (approved):
The next step in the right direction. Still more fixes to be done, but this gets us closer. Approved.

danielhvs (approved):
I've pushed a bugfix commit. Now LGTM even though we are still missing sending the new arg in the clj/get-layers request. Feel free to open a new PR for it so we can merge this one.

image

danielhvs:
please open following up PRs.

drewverlee:
@danielhvs me and @aliceliquori tested and watched it send the query params, i think your change to the arraymap, while potential fixing one issue, might have caused another. Can you please not make changes to our branch and then commit it? it makes it hard to understand whats going on.

drewverlee:
@danielhvs did you remove the weather-params from the config.cljs? It needed to be there! Thats what was allowing it to be sent in the query params.

src/cljs/pyregence/config.cljs
```diff
@@ -329,61 +329,58 @@
                                                               [:br]
                                                               [:strong "Firebrand Ignition Probability"]
                                                               " - An estimate of the probability that a burning ember could ignite a receptive fuel bed based on its temperature and moisture content."]
-                                                 :options
-                                                 (merge
-                                                  (when (nfdrs?) nfdrs-weather-params)
```
drewverlee, line 334: @danielhvs please don't change our branch code. Please just point out of if the feature doesn't work. I think you're saying that this needs to be an array map for the Weather params to be sorted the way they were before. 

We can probably do that and keep the necessary merge of weather-params, so they show up in the query to the server...

danielhvs:
> @danielhvs did you remove the weather-params from the config.cljs? It needed to be there! Thats what was allowing it to be sent in the query params.

It was not working for me. Please test again to double-check. Maybe you were in the middle of a REPL session with the data in place... IDK

src/cljs/pyregence/config.cljs
danielhvs: Sorry for that. Let's make it as a process then: I'll never push changes the PRs, just make suggestions. As we talked, we need to agree on the process as a team. From my side from now on I'm just going to comment in PRs. The idea was just to contribute but sorry again if this has caused confusion.

PR #1115, opened 2026-06-30
Revert "PYR-1555/Send nfdrs weather param query to server to get layers (#1114)"

danielhvs:
Reverts the squash-merge of #1114 (`PYR-1555/Send nfdrs weather param query to server to get layers`).

The author requested a revert. This PR reverts the merged changes from `main` back to the state at `9fcf2d3b`.

The original changes are not lost — I've restored the original PR commits on the **`send-nfdrs-weather-param-query-to-server-to-get-layers`** branch, so they remain available there for re-work / re-submission.

### DIFF
--- a/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
+++ b/src/cljs/pyregence/components/map_controls/collapsible_panel.cljs
@@ -339,9 +339,50 @@
                              #(select-param! % key)
                              selected-param-set]])))
                      (cond-> @!/processed-params
-                       (c/nfdrs?)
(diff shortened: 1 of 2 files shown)


danielhvs:
@drewverlee please give the +1 so that we can go back to original `main` state. 
Then please open a PR off `send-nfdrs-weather-param-query-to-server-to-get-layers`
Sorry again for the inconvinience.

drewverlee:
Thanks for trying, but this doesn't bring the repo back to the state where it needed to be for the nfdrs query param to be sent to the server. Look at the change in the two commits prior to yours. I'll create another branch and handle it there.

PR #1118, opened 2026-07-01
PYR1-1512 Enforce per-resource authorization on org/user and match-drop endpoints

danielhvs:
## Summary

Fixes the IDOR (Insecure Direct Object Reference) exposure behind PYR1-1512 by enforcing **per-resource authorization** on the org/user admin and match-drop endpoints. Following the [OWASP IDOR Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Insecure_Direct_Object_Reference_Prevention_Cheat_Sheet.html), the fix is an access-control check scoped to the acting user (derived from the session), not obfuscation of identifiers.

The endpoints were **role-gated but not resource-scoped**: the route layer verified a caller was *an* `organization_admin`, but the handlers then acted on whatever org/user id the client supplied. So an admin of one organization could edit another org, or change/remove users in it, by passing a foreign id. Match-drop delete/status acted on any job id regardless of owner.

## Changes

- **New SQL authorization predicates** (`organization_functions.sql`):
  - `can_admin_org(requesting_user, org_id)` — `super_admin`/`account_manager` may administer any org; an `organization_admin` only their own.
  - `can_admin_user(requesting_user, target_user)` — delegates to `can_admin_org` via the target user's org.
- **Gated the org/user admin handlers** (`authentication.clj`) on those predicates, using the acting user from the session; cross-org/cross-user requests now return `403`:
  - `update-org-info`, `add-org-user` (org-scoped)
  - `remove-org-user`, `update-org-user-role`, `update-user-org-membership-status`, `update-user-name` (user-scoped)
  - `super_admin`/`account_manager` remain unrestricted.
- **Scoped match-drop to the owner** (`match_drop.clj`): `delete-match-drop!` and `get-md-status` verify the job belongs to the session user, returning a uniform `403` that does not reveal whether the job exists.
- **Tests**: added `^:rct/test` blocks asserting the predicates allow same-org and deny cross-org access, that each guarded handler returns `403` across org boundaries, and that match-drop access to a non-owned job is denied.

No schema or primary-key changes: existing integer ids are kept; the fix is purely authorization. The predicates are `CREATE OR REPLACE` functions loaded by `bb functions` on deploy.

## Test plan

- [x] `bb test` — 22 tests, 129 assertions, 0 failures.
- [ ] As an `organization_admin`, confirm actions on your own org/users still succeed and actions targeting another org/user return `403`.
- [ ] Confirm `account_manager`/`super_admin` retain cross-org access.
- [ ] Confirm a user cannot view (`get-md-status`) or delete another user's match drop.

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -688,10 +688,12 @@
 
 (defn update-user-name
   "Allows an organization admin to update the name of a user by their email."
-  [_ email new-name]
+  [{:keys [user-id]} email new-name]
   (if-let [user-id-to-update (sql-primitive (call-sql "get_user_id_by_email" email))]
-    (do (call-sql "update_user_name" user-id-to-update new-name)
-        (data-response (str "User's name successfully updated to " new-name)))
+    (if (sql-primitive (call-sql "can_admin_user" user-id user-id-to-update))
+      (do (call-sql "update_user_name" user-id-to-update new-name)
+          (data-response (str "User's name successfully updated to " new-name)))
+      (data-response "You are not authorized to manage this user." {:status 403}))
     (data-response (str "There is no user with the email " email)
                    {:status 403})))
 
@@ -792,13 +794,15 @@
                 :archived-date         archived_date}))
        (data-response)))
 
-(defn add-org-user [_ org-id email]
-  (if-let [user-id-to-add (sql-primitive (call-sql "get_user_id_by_email" email))]
-    (do
-      (call-sql "add_org_user" org-id user-id-to-add)
-      (data-response ""))
-    (data-response (str "There is no user with the email " email)
-                   {:status 403})))
+(defn add-org-user [{:keys [user-id]} org-id email]
+  (if-not (sql-primitive (call-sql "can_admin_org" user-id org-id))
+    (data-response "You are not authorized to manage this organization." {:status 403})
+    (if-let [user-id-to-add (sql-primitive (call-sql "get_user_id_by_email" email))]
+      (do
+        (call-sql "add_org_user" org-id user-id-to-add)
+        (data-response ""))
+      (data-response (str "There is no user with the email " email)
+                     {:status 403}))))
 
 (defn add-org-users [{:keys [user-role organization-id]} org-id users]
   ;; Prevent none SA&AM from changing other orgs.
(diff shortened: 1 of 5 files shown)


danielhvs:
After this pr is merged a couple of other ones will be created for the ticket

src/clj/pyregence/authentication.clj
```diff
@@ -900,21 +883,11 @@
...
-  (data-response ""))
-
-(defn update-org-user-role [_ user-id new-role]
-  (call-sql "update_org_user_role" user-id new-role)
-  (data-response ""))
-
-(defn update-user-org-membership-status [_ user-id new-status]
-  (call-sql "update_org_membership_status" user-id new-status)
-  (data-response ""))
+(defn update-org-info [{:keys [user-id]} org-id org-name email-domains auto-add? auto-accept?]
```
drewverlee, line 886: This change looks right, we should always pull the user-id from the session!!! What was happening before???

drewverlee: @danielhvs, we also need to pull the org-id from the session, unless this api end-point allows a role like SA to do something cross-organization.

src/sql/functions/organization_functions.sql
```diff
@@ -1,6 +1,26 @@
...
 
+--------------------------------------------------------------------------------
+---  Authorization predicates
+--------------------------------------------------------------------------------
+-- True when the requesting user may administer the given organization.
+-- super_admins and account_managers may administer any org; an
+-- organization_admin may administer only their own org. Used to prevent IDOR:
+-- org-scoped mutations must gate on this so an admin of one org cannot act on
+-- another org by passing its id.
+CREATE OR REPLACE FUNCTION can_admin_org(_requesting_user_id integer, _org_id integer)
```
drewverlee, line 12: great idea, i have sql code which does this check, is there a way to compose sql using functions... i dont think so but worth brainstorming.

src/clj/pyregence/match_drop.clj
```diff
@@ -462,16 +462,23 @@
...
-   via the 'remove' action passed to the GeoSync Microservice"
-  [_ match-job-id]
-  (let [sig3-endpoint (get-config :triangulum.views/client-keys :features :sig3-endpoint)]
-    (delete-match-drop-using-kubernetes! sig3-endpoint match-job-id)))
+   via the 'remove' action passed to the GeoSync Microservice.
+   Only the match drop's owner may delete it."
+  [{:keys [user-id]} match-job-id]
+  (let [job (get-match-job-from-match-job-id! match-job-id)]
+    (if (and job (= user-id (:user-id job)))
+      (let [sig3-endpoint (get-config :triangulum.views/client-keys :features :sig3-endpoint)]
```
drewverlee, line 470: yes, like you do here we need to check.

src/clj/pyregence/match_drop.clj
```diff
@@ -462,16 +462,23 @@
...
 (defn get-md-status
-  "Returns the current status of the given match drop run."
-  [_ match-job-id]
-  (data-response (-> (get-match-job-from-match-job-id! match-job-id)
-                     (select-keys [:display-name :geoserver-workspace :message :md-status :job-log]))))
+  "Returns the current status of the given match drop run.
+   Only the match drop's owner may view it."
+  [{:keys [user-id]} match-job-id]
+  (let [job (get-match-job-from-match-job-id! match-job-id)]
+    (if (and job (= user-id (:user-id job)))
```
drewverlee, line 479: 👍

src/clj/pyregence/match_drop.clj
```diff
@@ -513,3 +520,15 @@
...
           (data-response {:type "FeatureCollection" :features []}
                          {:type :json})))))
+
+^:rct/test
+(comment
+  ;; --- Match drop ownership guard (PYR1-1512) ---
+  ;; A match drop the session user does not own (here a non-existent job) must
+  ;; not be viewable or deletable; both return 403 without revealing existence.
+  (get-md-status {:user-id 1} 999999999)
+  ;=>> {:status 403}
```
drewverlee, line 530: 👍

drewverlee (commented):
At a glance this looks good, thanks @danielhvs 

How does this solve the wider issue of not showing the user-id though? The user can still get the user-id through the session? 

However, i don't think thats an issue as long as the id (be it email or uuid) is IN the session, because the user can't tamper with it. However, a pendantic security aduit could claim that using an id (or to a far lesser extent email) still reveals information.

However were an open source project, so we dont' have those secrets anyway?

Regardless this looks like a good change.

danielhvs:
> How does this solve the wider issue of not showing the user-id though? The user can still get the user-id through the session?

I'll open 4 more PRs: each one will tackle each entity that exposes the PK, @drewverlee

src/sql/functions/organization_functions.sql
danielhvs: Yes. Like so:
```sql
  CREATE OR REPLACE FUNCTION update_org_info(_requesting_user_id integer, _org_id integer, ...)
  RETURNS void AS $$
      UPDATE organizations
      SET org_name = _org_name
      WHERE organization_uid = _org_id
        AND can_admin_org(_requesting_user_id, _org_id);  -- reuse the predicate
  $$ LANGUAGE SQL;
```

src/clj/pyregence/authentication.clj
danielhvs: Good question, Drew. After checking I think adding it would actually break the feature we have:

Analysis (from AI with me asking stuff etc.):
An `account_manager` has **no org** in the session (`organization = 'none'`). Their whole job is managing orgs they don't belong to.

```clojure
;; Session-derived org-id:
(update-org-info {:user-id 3 :organization-id nil} ...)  ;; AM, no org
;; → tries to update org nil → account manager can't do their job
```

Meanwhile the client sent `org-id 5` because that's the org they're actually editing in the UI. Ignoring it and using the session breaks the exact role the endpoint exists to serve — while `can_admin_org` already blocks an org-admin from touching org 5 if it isn't theirs.

sifbiri (approved):

PR #1121, opened 2026-07-02
PYR1-1108 Password needs min 12 chars max 128 one uppercase and one number

drewverlee:
As the title and ticket says, the server needs to verify passwords are between 12 and 128 chars with one upper case and one number. 

@sifbiri this is a draft atm beause i want to quickly check with you that the verify password functionality should only be attached in two places `add-new-user` and `set-user-password`

Also, can you glance at my regex and verify password logic? Does it have any issues? this can be subtle sometimes so i want a couple quick eyes on it.

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -183,8 +183,6 @@
 (defn set-user-password
   "Sets a new password for user with valid reset token."
   [session email password token]
-  ;;TODO should this be a cond? (so we dont nest ifs)
-  ;;TODO should this share more logic with add-new-user when it comes to passwords?
   (if-not (valid-password? password)
     (data-response (password->invalid-password-msg password) {:status 400})
     (if-let [user (first (call-sql "set_user_password" {:log? false} email password token))]
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -60,6 +60,50 @@
          _     (.nextBytes (SecureRandom.) bytes)]
      (.encodeToString (Base64/getEncoder) bytes))))
 
+(defn valid-password?
+  "Passwords must be between 12 and 128 chars and contain at least 1 upper case letter and number."
+  [password]
+  (and
+   ;; between 12 and 128
+   (<= 12 (count password) 128)
+   ;; digit
+   (boolean (re-find #"\d" password))
+   ;; upper case
+   (boolean (re-find #"[A-Z]" password))))
+
+^:rct/test
+(comment
+  ;; Must have 12 characters
+  (valid-password? "Helloworld1")
+  ;; => false at 11
+
+  (valid-password? "Helloworld12")
+  ;; => true at 12
+
+  (valid-password? (clojure.string/join "" (repeat 11 "Helloworld12")))
+  ;; => false over 128
+
+  (valid-password? (clojure.string/join "" (repeat 10 "Helloworld12")))
+  ;; => true less then 128
+
+;; Must have one upper case
+  (valid-password? "helloworld12")
+  ;; => false with no upper case
+
+  (valid-password? "Helloworld12")
+  ;; => true with one
+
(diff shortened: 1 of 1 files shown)


src/clj/pyregence/authentication.clj
```diff
@@ -60,6 +60,46 @@
          _     (.nextBytes (SecureRandom.) bytes)]
      (.encodeToString (Base64/getEncoder) bytes))))
 
+(defn valid-password?
+  "Passwords must be between 12 and 128 chars and contain at least 1 upper case letter and number."
```
kbeehler-sig: "Your password must be between 12 and 128 characters long, contain at least one number (0-9), and use a mix of uppercase and lowercase letters."

sifbiri:
Thanks, Drew.

Two places looks right to me - set_user_password and add_new_user are the only spots a user picks a password. add-org-users generates its own, no check needed.

The 400 echoes the raw password ("Invalid Password:'short'...") to the browser and the log, you can drop it.

bcrypt only reads the first 72 bytes, so the 128 max is misleading, you can cap at 72 (matches the OWASP page in the ticket).

Non-string input throws instead of returning false - maybe put (string? password) first.

drewverlee:
@sifbiri should we be logging raw passwords? Now we have passwords in our logs, which means they're easier to steal. What do we gain from having them recorded?

Will want a message telling clients (and the front end) what to do. Though it might take a different from on the browser client. 

Likely ill be turning this into a spec soon. thanks for the review.

src/clj/pyregence/authentication.clj
```diff
@@ -60,6 +60,52 @@
          _     (.nextBytes (SecureRandom.) bytes)]
      (.encodeToString (Base64/getEncoder) bytes))))
 
+(defn valid-password?
+  "Passwords must be between 12 and 64 chars and contain at least 1 upper case letter and number."
```
kbeehler-sig, line 64: @drewverlee  can we change the message to this:

Your password must be between 12 and 64 characters long, contain at least one number (0-9), and use a mix of uppercase and lowercase letters.

src/clj/pyregence/authentication.clj
```diff
@@ -60,6 +60,52 @@
...
+  (valid-password? "Helloworldaa")
+  ;; => false with no number
+
+  (valid-password? "Helloworld12")
+  ;; => true with at least one number
+  )
+
+(defn password->invalid-password-msg
+  [password]
+  (str "Invalid Password:'" password "'. Passwords must be between 12 and 64 chars and contain at least 1 upper case letter and number. "))
```
kbeehler-sig, line 107: Your password must be between 12 and 64 characters long, contain at least one number (0-9), and use a mix of uppercase and lowercase letters.

src/clj/pyregence/authentication.clj
drewverlee: @kbeehler-sig that string isn't shown to the user. It's just for documentation.

src/clj/pyregence/authentication.clj
drewverlee: > and use a mix of uppercase and lowercase letters.

We have to be specific. How many of each? Lets say at least one i guess.

kbeehler-sig: The message implies that it will be at least one of each because you're using a mix. This is what I'd like for the user facing phrasing.

src/clj/pyregence/authentication.clj
kbeehler-sig: Awesome. Resolved then.

src/clj/pyregence/authentication.clj
drewverlee: @kbeehler-sig will do. I'm updating another branch that has these changes fwiw. 

This string is only customer-facing in that if someone hits the api direclty, without using the client, they would see this. I'm not sure we even want to support signing up without using the browser client so it's kinda just there just in case.

PR #1122, opened 2026-07-06
PYR1-1559  Enable Secure session cookies and HSTS. Bump triangulum

sifbiri:
## Purpose
Bumps triangulum and turns on its HTTPS hardening: Secure session cookies, HSTS, and Server-header suppression.

Staging and prod need the same two keys in their deployed `config.edn` to take effect. Plain-HTTP dev should set `:secure false`.


danielhvs (approved):
LGTM

PR #1123, opened 2026-07-06
PYR1-1512 Expose match jobs by unpredictable uuid instead of sequential id

danielhvs:
## Purpose
Stop exposing the sequential match-job integer PK to the browser and use an unpredictable uuid as the public handle for match drops (PYR1-1512 enumeration hardening).

- Add a `match_job_uuid` (`uuid`, `gen_random_uuid()`) public identifier to `match_jobs` and use it as the browser-facing handle; the sequential integer PK stays internal.
- SQL: `get_match_job` / `get_user_match_jobs` return the uuid; add `get_match_job_by_uuid` for inbound lookups; internal functions keep the int PK. `get_user_match_jobs` no longer selects the sequential PK at all.
- Handlers: `get-match-drops` returns the uuid; `get-md-status` and `delete-match-drop!` accept the uuid and resolve it server-side, rejecting malformed ids with a uniform 403; `create-match-job` returns the uuid.
- Add a reusable `pyregence.utils/->uuid` coercion helper (nil on invalid input) and route match-drop coercion through it.
- cljs: dashboard and match-drop tool use `match-job-unique-id`; the dashboard sorts by creation time since the uuid is not ordinally sortable. Match-drop email reads `match-job-unique-id`.

Known limitation (documented in-code as accepted low-severity risk): the PK is still indirectly visible via the GeoServer workspace name (`match-drop-forecast_md-<pk>_...`) returned by `get-md-status` / `get-match-drops`. It is owner-scoped and only reveals system job volume; a full fix requires renaming the sig3/GeoServer workspace and is out of scope here.

## Related Issues
Closes PYR1-1512

## Submission Checklist

#### Module Impacted
Match Drop > Dashboard, Match Drop Tool

#### Role
User

#### Steps
1. Submit a match drop and open the Match Drop dashboard.
2. Confirm the job list loads, status polling works, and the job can be deleted.
3. Inspect the network responses: the browser-facing payloads carry the `match_job_uuid`, not the sequential integer id.
4. Call `get-md-status` / `delete-match-drop` with a malformed or foreign id and confirm a uniform 403 (no existence leak, no 500).

#### Desired Outcome
Match drops are addressed by their uuid end to end; the sequential PK never leaves the DB for browser-facing paths; malformed/foreign ids are rejected with a uniform 403.

---

## Screenshots
N/A (no visual changes)

### DIFF
--- a/src/clj/pyregence/email.clj
+++ b/src/clj/pyregence/email.clj
@@ -122,7 +122,7 @@
                     :failed    "Match Drop Failed"
                     "Match Drop Update")
         body      (messages/match-drop-email fmt base-url email user-name
-                                             (:match-job-id match-drop-args)
+                                             (:match-job-unique-id match-drop-args)
                                              (:display-name match-drop-args)
                                              status)
         result    (send-mail email nil "PyreCast Support" subject body fmt)]
--- a/src/clj/pyregence/match_drop.clj
+++ b/src/clj/pyregence/match_drop.clj
@@ -119,7 +119,7 @@
   (-> result
       (rename-keys {:match_job_id          :match-job-id
                     :match_job_uuid        :match-job-unique-id
-                    :runway_job_id         :runway-job-id
+                    :sig3_job_id           :sig3-job-id
                     :user_id               :user-id
                     :created_at            :created-at
                     :updated_at            :updated-at
@@ -165,7 +165,7 @@
   "Updates any of the properties of a match job based on a match-job-id (required).
    Any keys that are not provided will not be updated in the DB."
   [{:keys [match-job-id
-           runway-job-id
+           sig3-job-id
            md-status
            display-name
            message
@@ -177,7 +177,7 @@
   {:pre [(some? match-job-id)]}
   (call-sql "update_match_job"
             match-job-id
-            runway-job-id
+            sig3-job-id
             md-status
             display-name
             message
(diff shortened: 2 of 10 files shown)


sifbiri:
Nice work, @danielhvs. Not sure if this will deploy cleanly though, as it stands `functions` currently runs before `migrate`. `match_job_uuid` isn't there yet when functions runs, and `get_match_job_by_uuid` plus the two changed functions fail to build. Might need to add the column, then drop and recreate all three functions in a migration.

danielhvs:
@sifbiri you were right that this would not deploy cleanly, and the root cause turned out to be the deploy order rather than this migration: `functions` runs before `migrate`, so the three functions selecting `match_job_uuid` fail against a column that does not exist yet. `clear_functions.sql` drops every function first, so they end up dropped rather than stale, and `psql` exits 0, so the deploy still reports success.

Fixing the order is enough on its own, so this PR keeps the migration to just the column and its index, with no drops and no duplicated function bodies.

**Depends on PyreCast/pyrecast-gcp-deployment#46, which must land first.** Merging and deploying this before that one leaves match drop with no match functions at all.

Verified by replaying all 10 migrations from a zero-migration baseline as migrate then functions: clean throughout.

I reviewed and analyzed this PR in collaboration with Claude Code.

sifbiri (approved):

sifbiri:
Nice @danielhvs.

danielhvs:
Thanks, Sif. 
I have 2 more PRs to go for PYR1-1512

PR #1126, opened 2026-07-08
PYR1-1548/add client side password strength validation and indicator

drewverlee:
Add password validation logic to the client and server.

Passwords are now validated to be between 12 and 64 chars and contain at least 1 upper case letter and number. This validation happens on the browser on account registration and password reset in the form of a validation checklist to help
the user create a valid password.

The server also validates the password, though, it might not be possible to use the API to set a password.

A further improvement could be checking how random the password is and consolidating logic between the front and backend using cljc.

Note, despite what the branch title and PYR1- says, this covers server and client, Jira tickets 1548 and 1108. This was done to make it easier to make sure these two played nice together.

Also Note this doesn't force old password to be updated.


src/clj/pyregence/authentication.clj
```diff
@@ -60,6 +60,53 @@
          _     (.nextBytes (SecureRandom.) bytes)]
      (.encodeToString (Base64/getEncoder) bytes))))
 
+;;TODO consider sharing this logic on the frontend
+(defn valid-password?
+  "Passwords must be between 12 and 64 chars and contain at least 1 upper case letter and number."
```
kbeehler-sig: @drewverlee : For consistency with my other comments on the server-side. Can we update the message to read:

Your password must be between 12 and 64 characters long, contain at least one number (0-9), and use a mix of uppercase and lowercase letters.

drewverlee: The message here isn't something the Users see, however, it's nice to have it close as possible to the wording so developers get the same idea. Thanks for the suggestion, I updated the code.

danielhvs (changes requested):
Nice work, Drew! I'm glad to se we're adding tests to it...

1. Remove plaintext password from server error message — authentication.clj:116 (used :198, :687). Reason: leaks the password into
  logs/error-tracking/history.
  2. Delete the dead "at least 8 characters" check — register.cljs:61, reset_password.cljs:39. Reason: superseded by the ≥12 rule;
  contradictory duplicate message.
  3. Add ^{:key text} to the for - validations.cljs:36. Reason: avoids React key warning / index-based reconciliation.
  4. Rename f → password-checks (or other name you like) (validations.cljs:6). Reason:  f is opaque; 
  5. drop ! from ->toast! (:19) and ->cmpt! (:30) fn names. Reason: ! wrongly implies
  side effects on pure fns.

src/cljs/pyregence/components/password/validations.cljs
```diff
@@ -0,0 +1,41 @@
...
+
+(defn- f
+  [password]
+  [{:validation/text   "One number."
+    :validation/valid? (boolean (re-find #"\d" password))}
+   {:validation/text   "One lowercase."
+    :validation/valid? (boolean (re-find #"[a-z]" password))}
+   {:validation/text   "One uppercase."
+    :validation/valid? (boolean (re-find #"[A-Z]" password))}
+   {:validation/text   "12 characters."
```
danielhvs:
```suggestion
   {:validation/text   "At least 12 characters."
```

danielhvs:
FE tested seems right: adds the feature in both new acc and resetting pwd. 

image

image

drewverlee:
Thanks daniel.

> Remove plaintext password from server error message — authentication.clj:116 (used :198, :687). Reason: leaks the password into logs/error-tracking/history.

Can we please make a ticket to address that the data-response function automatically logs information when all it implies is that it sends a data-response?

drewverlee:
@danielhvs I made the suggested changes.

I renamed f to data, but keep in mind the full name was password.validations/f and is now password.validations/data. That function is a set of validations, but i didn't want to shadow the word set, i think if done properly, every namespace will have at least one very vague function or ref because it's at the heart of the set of functions in that namspace and so represented largely by the namespace name itself.

danielhvs:
> Thanks daniel.
> 
> > Remove plaintext password from server error message — authentication.clj:116 (used :198, :687). Reason: leaks the password into logs/error-tracking/history.
> 
> Can we please make a ticket to address that the data-response function automatically logs information when all it implies is that it sends a data-response?

Yes. [here](https://sig-gis.atlassian.net/browse/PYR1-1569) it is. (Sif has created it)

danielhvs (changes requested):
- #2 ❌ - the dead "Your password must be at least 8 characters long." check is still live at register.cljs:61 and
  reset_password.cljs:39. It's superseded by the ≥12 rule in ->toast, so on a short password the toast will list both the stale 8-char
  message and the real "at least 12 characters" one. Low severity, but it's the one item left unaddressed.

danielhvs (approved):

PR #1132, opened 2026-07-17
PYR1-1560/fire-popup-shows-sources-more-clearly

drewverlee:
Now when you click an active fire the pop-up shows the sources more clearly. Specifically Pyrecast is cited in the "view forecast" section and Watchduty is cited as the initial burn source location. This should look like the figma designs in https://www.figma.com/design/QitY9QZbsGqFL1OuUZDKsG/Pyrecast?node-id=1528-4237&t=LuDSClfMWB3LWfon-0

Watch Duty Options - Iteration 5, slide Option 7 - Final Option

Here is how it looks in this branch:

image


danielhvs (changes requested):
I reviewed and analyzed this PR, in collaboration with Claude Code. Nice cleanup - the layout matches the Figma well. Three small things I'd like changed before merge:

1. **Divider renders unconditionally** (`popups.cljs:87`). Both `show-link?` and `source` are optional, so when either is missing you get a dangling 2px gray line (plus the parent's 16px gap on each side). Suggest `(when (and show-link? source) [:div ...])`.

2. **Hardcoded `#E1E1E1`** (`popups.cljs:87`). `($/color-picker :neutral-soft-gray)` is `rgba(225, 225, 225)` - an exact match, and consistent with the rest of the file.

3. **Redundant wrapper in `fire-link`** (`popups.cljs:62`). The outer `[:div]` no longer carries any style; the button can be returned directly. There's also a stray blank line at line 63.

drewverlee:
@danielhvs changes made thanks for noticing those, especially the divider one!

danielhvs (approved):
LGTM

PR #1134, opened 2026-07-17
PYR1-1511 Session security: invalidation, timeouts, single session

sifbiri:
## Purpose

Logout didn't really revoke access. Sessions are stateless cookies with no server-side record, so once one was issued the server couldn't kill it - a copied cookie outlived logout, sessions never expired, and one account could stay logged in everywhere. Pentest findings, under epic PYR1-1511.

The fix lets the server reject a dead session. `route-authenticator` checks two things: a per-user cutoff (`users.session_invalidated_at`, epoch ms) that rejects any session older than it, and `:created-at` / `:last-active` stamps against idle and absolute timeouts. Logout and login move the cutoff to now. Only login-required routes are gated, so login, logout, reset and 2FA stay reachable with a stale cookie.

- 1513 - logout moves the cutoff, so a replayed pre-logout cookie stops working (log-out-everywhere for free).
- 1514 - logout also clears the cookie and sends `Clear-Site-Data`. Same acceptance criteria as 1513.
- 1515 - sessions time out at 15 min idle / 7 h absolute (configurable, inside the AC). Activity keeps the idle clock alive; an expired session is never refreshed.
- 1519 - a new login drops the older sessions. A strict `<` lets the fresh login survive its own bump.
- 1562 - the org endpoint was token-only and leaked a logged-out cookie its org data and any geoserver credentials. Now it needs a live session.

Also points a dead session to `/login`, extracts an `auth-types` helper, and folds the four subscription-tier cases into one.

`bb test`, `bb test-clj`, and clj-kondo are green, with a live run against a real database.

Deploy: `bb migrate` (column) then `bb functions` (SQL); `bb deploy` runs functions, a fresh build needs neither. 

**Heads-up:** everyone is logged out once at deploy, since older cookies have no timestamps and fail closed.

### DIFF
--- a/src/clj/pyregence/handlers.clj
+++ b/src/clj/pyregence/handlers.clj
@@ -35,14 +35,6 @@
 
 (def not-found-handler (comp #(assoc % :status 404) (render-page "/not-found")))
 
-(defn redirect-handler [{:keys [session query-string uri] :as _request}]
-  (let [full-url (url-encode (str uri (when query-string (str "?" query-string))))]
-    (if (:user-id session)
-      (redirect (str "/?flash_message=You do not have permission to access "
-                     full-url))
-      (redirect (str "/login?flash_message=You must login to see "
-                     full-url)))))
-
 (def role-hierarchy
   (-> (make-hierarchy)
       (derive :super-admin         :organization-admin)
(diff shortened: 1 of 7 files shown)


danielhvs:
## Manual verification of PYR1-1513 (logout invalidation)

**TL:DR: PYR1-1513 LGTM**

Ran the ticket's repro against this branch locally (`4b6aef56`, `http://localhost:8080`).

**Probe route:** `POST /clj/get-current-user-settings` (`:auth-type :member`, `src/clj/pyregence/routing.clj:90`) - deliberately *not* the ticket's `/account-settings`, see note below.

```bash
C='ring-session=<value copied from Firefox DevTools -> Storage -> Cookies>'

curl -s -w '\nstatus=%{http_code}\n' -X POST \
  http://localhost:8080/clj/get-current-user-settings \
  -H 'Content-Type: application/edn' -H "Cookie: $C" -d ''
```

| # | Condition | Result |
|---|---|---|
| 1 | No cookie (control) | `403` `"User not found"` |
| 2 | Corrupt cookie (control) | `500` - signature failure, distinct path |
| 3 | Valid cookie, **before** logout | `200` `{:settings "{:timezone :utc}", :email "admin@all-orgs.com"}` |
| 4 | Same cookie, **after** logout | `403` `"Forbidden"` |

**Verdict:** 3 -> 4 on an identical cookie is the acceptance criterion. The token was live, logout revoked it server-side, and replay is refused. The distinct `"Forbidden"` (vs. `"User not found"` in case 1) confirms the cookie still decrypts - it's the session lookup rejecting it, not a crypto failure, which is the correct mechanism.

### Worth fixing on the ticket

PYR1-1513 step 5 uses `GET /account-settings` as the probe, which is not a valid test here: it returns `200` with a byte-identical 3122-byte SPA shell whether authenticated or not (diffed both bodies). Anyone re-running the ticket as written gets a false positive. Step 5 should target an auth-enforcing `/clj/` route instead.

### Caveat

This exercises logout revocation only. The idle/absolute timeouts and login-time rotation also in this PR were not covered by this run.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

danielhvs:
## PYR1-1514: the logout `Set-Cookie` does not clear the cookie

**Tested on** `4b6aef5`, local `http://localhost:8080`, config without `session-cookie-attrs`. Captured a live `ring-session` from Firefox, called `POST /clj/log-out` with the session cookie + bearer token, and replayed the cookie against `POST /clj/get-current-user-settings` (`:auth-type :member`).

Server-side revocation is fine: `200` + user data before logout, `403 "Forbidden"` after. Client-side is not:

```
Clear-Site-Data: "cache", "cookies", "storage"
Set-Cookie: ring-session=J8dMIfrA9wsjrXwWFK3FQWi7GPB9pWYkL8UiOe6juQU%3D--...;Path=/;HttpOnly
```

No `Max-Age=0`, no past `Expires`. Root cause is `ring.middleware.session.cookie:102-103` - `delete-session` is `(seal secret-key {} options)`, so for a `cookie-store` a "delete" re-seals an empty map and Ring emits it as the cookie *value*. `{:session nil}` can never produce an expiry.

So cookie removal rests entirely on `Clear-Site-Data`, which Safari ignores outright and which is dropped on non-secure origins - e.g. the `http://PyreCast.org` target in the ticket's own repro.

### Proposed fix (`authentication.clj:202`)

```clojure
(-> (data-response "" {:session nil})
    (assoc :session-cookie-attrs {:max-age 0})
    (assoc-in [:headers "Clear-Site-Data"] "\"cache\", \"cookies\", \"storage\""))
```

Simulated against Ring 1.9.6 with a real session key:

```
CURRENT  : ring-session=...;HttpOnly;Path=/
PROPOSED : ring-session=...;HttpOnly;Path=/;Max-Age=0
```

Notes: `:session-cookie-attrs` must be `assoc`ed onto the response, not passed in the params map - `utils/data-response` destructures only `:status :type :session` and drops the rest. A hand-written `Set-Cookie` header would be overwritten by `wrap-session`. `clj-handler` passes the response through untouched when `:session` is already set, so no interaction.

### Separately

`config.edn` is gitignored, so `config.default.edn:51` (`{:secure true :same-site :lax}`) is intent, not proof. Worth confirming the deployed cookie actually carries `Secure`/`SameSite` - if not, that's a bigger issue than this one.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

danielhvs:
## PYR1-1515 verified - idle and absolute timeouts work as specified

Manually tested in Firefox against `4b6aef5`, local server, with `:pyregence.auth/idle-timeout-min 1` and `:pyregence.auth/absolute-timeout-min 3` so both paths are observable in real time. Three scenarios, all passing:

**A - idle expiry.** Logged in, sat completely idle for 90s, then triggered a `/clj/` call. Session rejected. The idle cutoff fires as intended.

**B - absolute cap.** Logged in fresh, clicked continuously every ~30s for 3.5 minutes. Session died at ~3 min *despite* constant activity - the absolute cap correctly overrides a refreshed idle timer. This is the scenario that distinguishes the two mechanisms, and it holds.

**C - the refresh mechanism.** Confirmed the `ring-session` cookie value rotates on each `/clj/` POST, i.e. `:last-active` is genuinely re-sealed per request rather than being stamped once at login.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

danielhvs:
## PYR1-1519 verified - a new login invalidates the previous session

**Tested on** `4b6aef5`, local `http://localhost:8080`, Firefox.

Ran the ticket's acceptance steps as written:

1. Window A - logged in, opened Account Settings.
2. Incognito Window B - logged in as the same user.
3. Window A - reloaded Account Settings.

**Result:** Window A was sent to the login page; its session had been dropped. Window B stayed functional. Newest login wins, which is the acceptance criterion.

### Note on the mechanism

Login stamps `set_user_session_invalidated_at` with `now`, and `session-invalidated?` compares with a strict `<` against `:created-at`. That strictness is load-bearing: the session that *causes* the bump is created at the same millisecond, so `<=` would log the user out of the window they just logged into. Worth preserving if that predicate is ever touched.

### Caveat

This run covers login-time rotation only. Logout revocation and the idle/absolute timeouts are covered in the earlier comments on this PR.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

danielhvs:
## PYR1-1562 verified - the org endpoint now requires a live session

**Route:** `POST /clj/get-current-user-organization`, `:auth-type` changed from `:token` to `#{:member :token}` (`src/clj/pyregence/routing.clj:129`).

**Tested on** `4b6aef5`, local `http://localhost:8080`, Firefox, observed via DevTools -> Network.

Account choice matters here: all three call sites select this route only for non-`super_admin` / `account_manager` roles, falling back to `get-all-organizations` otherwise (`near_term_forecast.cljs:902`, `account_settings.cljs:38`, `organizations_utils.cljs:11`). Tested as `user@pyr.dev` (`organization_member`), the lowest-privilege account that actually exercises the change.

| # | Condition | Result |
|---|---|---|
| 1 | Live member session, `/account-settings` and `/` | `200`, org list returned as before |
| 2 | Anonymous (incognito, no login), `/` | `403` `"No user is logged in."` |

Case 2 is worth spelling out, because both outcomes are `403` and the **body** is the signal, not the status code. `"No user is logged in."` is the handler's own response (`authentication.clj`), which means the request passed the authenticator and reached the handler - `session-user-role` defaults to `:member` when there is no session, so an anonymous visitor is still admitted and the public map keeps loading. A `"Forbidden"` body there would have meant the route itself began blocking anonymous traffic, breaking `/` for logged-out visitors. It does not.

Rejection of a *dead* session on this route travels the same `route-authenticator` path already verified against `/clj/get-current-user-settings` in the PYR1-1513 and PYR1-1515 comments above, so it is not re-tested here.

### Robustness nit (not a blocker)

Two callers read the response without checking `:success`:

- `src/cljs/pyregence/pages/account_settings.cljs:35`
- `src/cljs/pyregence/pages/near_term_forecast.cljs:919`

Both pipe `(edn/read-string (:body ...))` straight into `!/user-orgs-list`. When this route returns `"Forbidden"` for a dead session, `edn/read-string` yields the symbol `Forbidden` rather than throwing, so a vector of maps is silently replaced by a symbol. Downstream consumers expect the vector (`forecast_tabs.cljs:47`, `wui.cljs:15`). `organizations_utils.cljs` does check `:success` and degrades to `{}`, so only those two are affected.

Not a security issue - the server-side rejection is correct - and invisible in dev data, since no seeded org is a PSPS/WUI org. Flagging so it is a known behaviour rather than a surprise later.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

danielhvs (changes requested):
## Review summary

I worked through the PR ticket by ticket against a local server on `4b6aef5`, with a live database. Results are in the five comments above.

| Ticket | Verdict |
|---|---|
| 1513 - logout revokes server-side | Verified |
| 1514 - logout clears the cookie | **Defect found** |
| 1515 - idle / absolute timeouts | Verified |
| 1519 - new login drops older sessions | Verified |
| 1562 - org endpoint requires a live session | Verified |

The session-invalidation core is solid. Four of the five hold up under manual testing, including the two mechanisms that are easiest to get subtly wrong - the absolute cap correctly overriding a refreshed idle timer, and the strict `<` that lets a fresh login survive its own invalidation bump.

### Requesting changes on one item: PYR1-1514

The logout `Set-Cookie` does not actually clear the cookie - no `Max-Age=0`, no past `Expires`. Ring's `cookie-store` implements "delete" as re-sealing an empty map, so `{:session nil}` can never produce an expiry. Cookie removal therefore rests entirely on `Clear-Site-Data`, which Safari ignores and which browsers drop on non-secure origins, including the `http://PyreCast.org` target named in the ticket's own repro.

Server-side revocation is unaffected and works, so this is a client-side residue problem rather than an access-control hole - but it is precisely the acceptance criterion 1514 is written against. One-line fix proposed in the comment above (`assoc :session-cookie-attrs {:max-age 0}`), verified against Ring 1.9.6.

Happy to re-check once that lands. Two non-blocking items also noted in the comments: the PYR1-1513 ticket text uses a probe route that cannot fail (worth correcting on the ticket, not in this PR), and two CLJS callers of the org endpoint parse the response without checking `:success`.

---

Reviewed and analyzed by me, in collaboration with Claude Code.

src/clj/pyregence/routing.clj
```diff
@@ -127,7 +127,7 @@
                                                      :auth-type #{:account-manager :token}
                                                      :auth-action :block}
    [:post "/clj/get-current-user-organization"]      {:handler (clj-handler authentication/get-current-user-organization)
-                                                     :auth-type :token
```
drewverlee, line 130: @sifbiri, is this change relevant to the ticket? If so how?

src/sql/functions/user_functions.sql
```diff
@@ -535,6 +535,21 @@ RETURNS timestamptz AS $$
     WHERE user_uid = _user_id;
 $$ LANGUAGE SQL;
 
+-- Mark all of a user's sessions invalidated as of _ts (epoch ms).
+CREATE OR REPLACE FUNCTION set_user_session_invalidated_at(_user_id integer, _ts bigint)
+RETURNS void AS $$
+    UPDATE users
+    SET session_invalidated_at = _ts
```
drewverlee, line 542: how does passing the time (_ts) here compare with calling PSQLs CURRENT_TIMESTAMP (and maybe having to do a conversion?). I don't have strong feelings besides it's sometimes nice to let the DB do more.

src/sql/functions/user_functions.sql
```diff
@@ -535,6 +535,21 @@ RETURNS timestamptz AS $$
     WHERE user_uid = _user_id;
 $$ LANGUAGE SQL;
 
+-- Mark all of a user's sessions invalidated as of _ts (epoch ms).
+CREATE OR REPLACE FUNCTION set_user_session_invalidated_at(_user_id integer, _ts bigint)
```
drewverlee, line 539: We could use user-email here instead of user-id, as i think long term we want to move away from passing around the database's user-id outside the db itself.

src/clj/pyregence/handlers.clj
```diff
@@ -95,50 +87,233 @@
...
+  [session now]
+  (session-expired? session now
+                    (timeout-ms :pyregence.auth/idle-timeout-min     default-idle-timeout-min)
+                    (timeout-ms :pyregence.auth/absolute-timeout-min default-absolute-timeout-min)))
+
+(defn- session-invalidated?
+  "Created strictly before the user's invalidation point (set on logout / newer login).
+   Strict `<` so a fresh login at the same instant survives; invalidated-at 0 = never."
+  [{:keys [user-id created-at]} invalidated-at]
+  (boolean (and user-id created-at (pos? invalidated-at) (< created-at invalidated-at))))
```
drewverlee, line 118: Why is it necessary to check anything besides that an invalidated-at number exists?

src/clj/pyregence/handlers.clj
```diff
@@ -95,50 +87,233 @@
...
-        ;; Extract Bearer token from Authorization header
         bearer-token           (some->> (get headers "authorization")
                                         (re-find #"(?i)^Bearer\s+(.+)$")
                                         second)
         valid-token?           (= bearer-token (get-config :triangulum.views/client-keys :auth-token))
         super-admin?           (isa? role-hierarchy user-role :super-admin)
         account-manager?       (isa? role-hierarchy user-role :account-manager)]
-    (every? (fn [auth-type]
-              (case auth-type
+    (every? (fn [one-auth-type]
```
drewverlee, line 202: why rename this to one-auth-type?

src/clj/pyregence/handlers.clj
```diff
@@ -95,50 +87,233 @@
...
-                                           account-manager?
-                                           (isa? subscription-hierarchy @subscription-tier :tier2-pro))
-                :tier1-basic-paid      (or super-admin?
-                                           account-manager?
-                                           (isa? subscription-hierarchy @subscription-tier :tier1-basic-paid))
-                :tier1-free-registered (or super-admin?
-                                           account-manager?
-                                           (isa? subscription-hierarchy @subscription-tier :tier1-free-registered))
+                ;; super-admins / account-managers have access to all tiers
+                (:tier3-enterprise :tier2-pro :tier1-basic-paid :tier1-free-registered)
```
drewverlee, line 218: nice use of the multiselect case!

drewverlee (commented):
Overall this looks good. Is it possible for me to test this locally? If so i'll do so tomorrow. Otherwise w can probably get this into dev and go from there.

src/clj/pyregence/routing.clj
danielhvs: Yes - this is PYR1-1562. `:token` gated only on the shared app bearer key (every browser sends it), and being token-only it skipped the liveness gate, so a logged-out cookie still got back org data + geoserver creds. Adding `:member` requires a live member session, closing the leak.

(I've reviewed your takes, in collaboration with Claude Code.)

src/sql/functions/user_functions.sql
danielhvs: I'd keep `_ts`. The column is epoch-ms `bigint`, compared in Clojure against the cookie's `:created-at`/`:last-active` (also `System/currentTimeMillis`). `CURRENT_TIMESTAMP` would mean a type conversion on every read and split the comparison across two clocks (DB vs app server). Passing `_ts` keeps a single clock source.

(I've reviewed your takes, in collaboration with Claude Code.)

src/sql/functions/user_functions.sql
danielhvs: Reasonable long-term, but I'd defer it: these are server-only calls using the session's user-id (the id never reaches the browser, so PYR1-1512 doesn't apply here), and every other function in this file is user-id-keyed. Better as a dedicated refactor than in this PR.

(I've reviewed your takes, in collaboration with Claude Code.)

src/clj/pyregence/handlers.clj
danielhvs: The comparison is the mechanism, not a guard. `session_invalidated_at` is a persistent per-user watermark, set on both logout **and** login and never cleared - so a bare existence check would log every user out on their first request after login. `< created-at invalidated-at` is what lets the new session survive while older ones die.

(I've reviewed your takes, in collaboration with Claude Code.)

src/clj/pyregence/handlers.clj
danielhvs: The outer `auth-type` is the route's whole requirement (keyword or set); the `every?` lambda binds one element. Reusing the same name shadowed it - `one-auth-type` disambiguates and made the `auth-types` helper extraction clean. No behavior change.

(I've reviewed your takes, in collaboration with Claude Code.)

danielhvs (approved):

PR #1135, opened 2026-07-24
PYR1-1207/add login throttling to pyrecast webapp and apis

drewverlee:
Now 6 invalid login attempts into an account will lock the user out until they reset the password via 'forgot password'.

Note this will need to be deployed alongside this [PR ](
                             )to update the deployment repo to have the pyregence config.edn changes to add:

```clojure
:triangulum.worker/workers
{:triangulum.worker/name  "reset-user-email->failed-login-attempts"
  :triangulum.worker/start pyregence.jobs/start-reset-user-email->failed-login-attempts!
  :triangulum.worker/stop  pyregence.jobs/stop-reset-user-email->failed-login-attempts!}
```

### DIFF
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -12,6 +12,17 @@
             [java.security SecureRandom]
             [java.text SimpleDateFormat]))
 
+;; TODO this will need a
+;; 4 attempts 0 time
+;; 2 attempts 10 second
+;; 2 30 second
+;; 2 90  exceptional
+;; 2 360 exceptional
+
+;TODO make a less dumb one tomorrow
+;;TODO will need to get reset
+(def user-email->failed-login-attempts (atom {}))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Helper Functions
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -125,18 +136,24 @@
    (marketplace/complete-signup! request-session user_email)
    (create-session-from-user-data user)))
 
+;;TODO this will have to store failed login attempts per user.
 (defn log-in
   "Authenticates user and determines 2FA requirements."
   [session email password]
-  (if-let [user (first (call-sql "verify_user_login" {:log? false} email password))]
-    (let [user-id    (:user_id user)
-          two-factor (:two-factor (get-user-settings user-id))]
-      (case two-factor
-        :totp  (data-response {:email email :require-2fa true :method "totp"})
-        :email (do (email/send-email! nil email :2fa)
-                   (data-response {:email email :require-2fa true :method "email"}))
-        (successful-login user session)))
-    (data-response "" {:status 403})))
+  (def args [session email password])
+  ;;TODO not nested if block
+  (if (< 2 (@user-email->failed-login-attempts email 0))
+    (data-response "" {:status 429})
+    (if-let [user (first (call-sql "verify_user_login" {:log? false} email password))]
+      (let [user-id    (:user_id user)
+            two-factor (:two-factor (get-user-settings user-id))]
+        (case two-factor
+          :totp  (data-response {:email email :require-2fa true :method "totp"})
+          :email (do (email/send-email! nil email :2fa)
+                     (data-response {:email email :require-2fa true :method "email"}))
+          (successful-login user session)))
+      (data-response "" {:status 403
+                         :failed-login-attempts (swap! user-email->failed-login-attempts update email (fnil inc 0))}))))
 
 (defn marketplace-sso-login
   "Marketplace SSO entry point. Validates JWT, auto-logs in or redirects to 2FA.
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -207,7 +207,9 @@
   (if-not (valid-password? password)
     (data-response password->invalid-password-msg  {:status 400})
     (if-let [user (first (call-sql "set_user_password" {:log? false} email password token))]
-      (successful-login user session)
+      (do
+        (swap! user-email->failed-login-attempts dissoc email)
+        (successful-login user session))
       (data-response "Invalid or expired reset token" {:status 403}))))
 
 (defn verify-user-email
--- a/src/clj/pyregence/authentication.clj
+++ b/src/clj/pyregence/authentication.clj
@@ -1,5 +1,6 @@
 (ns pyregence.authentication
-  (:require [pyregence.email            :as email]
+  (:require [clojure.string             :as str]
+            [pyregence.email            :as email]
             [pyregence.marketplace      :as marketplace]
             [pyregence.totp             :as totp]
             [pyregence.utils            :refer [nil-on-error ->uuid]]
@@ -13,6 +14,12 @@
             [java.text SimpleDateFormat]))
 
 (defonce user-email->failed-login-attempts (atom {}))
+
+(defn- normalize-email
+  "Lower-cases and trims an email so throttle lookups match the identity the
+   DB authenticates against (verify_user_login uses lower_trim)."
+  [email]
+  (-> email str/lower-case str/trim))
 
 ;;TODO As an improvement, this could be made to be user-email specific
 (defn reset-user-email->failed-login-attempts!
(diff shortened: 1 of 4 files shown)


aliceliquori (approved):
Great review session! ++

kbeehler-sig:
@drewverlee : Can't speak to the code here but I recommend changing the approach slightly so as to not display the countdown timer on the screen to be user. We should implement the throttling on the server side but not expose it on the UI explicitly. Continue using the existing invalid login credentials generic messaging when the user passes that through and keep displaying the forgot password link directly underneath the password input on the screen. Should the user lock themselves out let's create a second ticket that would deliver the email directly to the user's account and ask them to reset their password due to suspicious activity essentially. If you can point me to the email text for reset password and I can take a look there too.

src/clj/pyregence/authentication.clj
```diff
@@ -125,18 +135,24 @@
...
-          two-factor (:two-factor (get-user-settings user-id))]
-      (case two-factor
-        :totp  (data-response {:email email :require-2fa true :method "totp"})
-        :email (do (email/send-email! nil email :2fa)
-                   (data-response {:email email :require-2fa true :method "email"}))
-        (successful-login user session)))
-    (data-response "" {:status 403})))
+  (let [failed-login-attempts  (@user-email->failed-login-attempts email 0)]
+    ;;TODO sync max max-failed-login-attempts value with client
+    (if (<= 6 failed-login-attempts)
```
danielhvs, line 156:
```suggestion
    (if (<= 5 failed-login-attempts)
```

src/clj/pyregence/authentication.clj
```diff
@@ -125,18 +135,24 @@
...
+      (data-response {:failed-login-attempts failed-login-attempts} {:status 429})
+      (if-let [user (first (call-sql "verify_user_login" {:log? false} email password))]
+        (let [user-id    (:user_id user)
+              two-factor (:two-factor (get-user-settings user-id))]
+          (case two-factor
+            :totp  (data-response {:email email :require-2fa true :method "totp"})
+            :email (do (email/send-email! nil email :2fa)
+                       (data-response {:email email :require-2fa true :method "email"}))
+            (successful-login user session)))
+        (data-response {:failed-login-attempts ((swap! user-email->failed-login-attempts update email (fnil inc 0)) email)}
```
danielhvs: Maybe would be better to control it by adding a database column in the `user` table?
This would be more solid than a clojure atom IMHO

src/clj/pyregence/authentication.clj
```diff
@@ -12,6 +12,16 @@
...
 
+(defonce user-email->failed-login-attempts (atom {}))
+
+;;TODO As an improvement, this could be made to be user-email specific
+(defn reset-user-email->failed-login-attempts!
+  []
+  (loop []
+    ;; 5 minutes
+    (Thread/sleep (* 1000 60 5))
+    (reset! user-email->failed-login-attempts {})))
```
danielhvs: If we want to go with clojure atom and not with the database column we should clear it up by user after 5min, and not the whole atom after this time.

src/clj/pyregence/authentication.clj
```diff
@@ -12,6 +12,16 @@
             [java.security SecureRandom]
             [java.text SimpleDateFormat]))
 
+(defonce user-email->failed-login-attempts (atom {}))
+
+;;TODO As an improvement, this could be made to be user-email specific
+(defn reset-user-email->failed-login-attempts!
+  []
+  (loop []
```
danielhvs: This is a "loop without recur"

src/clj/pyregence/jobs.clj
```diff
@@ -59,3 +60,12 @@
 
 (defn stop-get-weather-stations! [fut]
   (future-cancel fut))
+
+
+(defn start-reset-user-email->failed-login-attempts! []
```
danielhvs, line 65: Add an entry in `config.default.edn` for this worker.

src/clj/pyregence/jobs.clj
```diff
@@ -59,3 +60,12 @@
...
+  (future
+    (log-str "Reset failed login attempt counter every five minutes.")
+    (reset-user-email->failed-login-attempts!)))
+
+(defn stop-reset-user-email->failed-login-attempts! [fut]
```
danielhvs, line 70: add this tear downing fn to config.default.edn (start up fn: https://github.com/pyregence/pyregence/pull/1135/changes#r3685395964)

src/cljs/pyregence/pages/login.cljs
```diff
@@ -28,21 +28,23 @@
 (defn- log-in! []
   (go
     ;;TODO consider validating email before sending it.
-    (let [response (<! (u-async/call-clj-async! "log-in" @email @password))]
-      (if (:success response)
-        (let [resp-data (edn/read-string (:body response))]
-          (if (:require-2fa resp-data)
+    (let [{:keys [success body]} (<! (u-async/call-clj-async! "log-in" @email @password))]
+      (let [{:keys [require-2fa email method failed-login-attempts] :as body} (edn/read-string body)]
```
danielhvs: `email` shadows pyregence.pages.login/email atom
to fix it:
1. join the 2 `let`
2. in the 2nd {:keys ....} association, do it like so:
```clojure
(let [{:keys [success body]} (<! (u-async/call-clj-async! "log-in" @email @password))
        {:keys [require-2fa method failed-login-attempts] resp-email :email} (edn/read-string body)]
      (if success
          ,,,
          ,,,))
```

danielhvs (changes requested):
Reviewed and analyzed this PR in collaboration with Claude Code. A few issues beyond my inline comments:

### Server / correctness

**1. Throttle bypass via email casing/whitespace.** The DB keys logins by `lower_trim(email)`, but `user-email->failed-login-attempts` is keyed by the raw client string. `Foo@bar.com`, `foo@bar.com`, and `" foo@bar.com "` are one account to the DB but three separate counters in the atom - so an attacker just varies case/whitespace to get a fresh 6-attempt budget each time against the same real account. Fix: normalize (`lower-case` + `trim`) the key before every read/`swap!`/`dissoc`. The DB-column approach I suggested inline would inherently fix this; the atom version does not.

**2. Successful login never clears the counter.** Reset only happens on password-reset, user-creation, and the (broken) global timer - **not** on a successful `successful-login`, and **not** on the verified `:totp`/`:email` 2FA branches (where the password *was* correct). A legit user who mistypes twice then logs in keeps a count of 2, so interspersed successes never give a clean slate. Recommend a `dissoc` on the success path (and the 2FA branches).

### Client robustness (several become moot if we drop the explicit lockout message per @kbeehler)

**3. Client "locked" message is off-by-one from the server.** The 6th failed attempt returns `403` with count 6, so the client shows "Account locked" one attempt before the server actually starts returning `429` (which first trips on the 7th request). This is also tied to my inline `6 -> 5` suggestion: the threshold lives in two places (server `log-in` and client `log-in!`), and the code's own TODO ("sync max-failed-login-attempts value with client") flags it - whatever number we pick has to be kept in sync on both sides. Moot if the UI message is removed per @kbeehler.

src/clj/pyregence/authentication.clj
drewverlee: and is that a problem?

src/clj/pyregence/authentication.clj
drewverlee: @danielhvs we can do that, but i'm not sure it enhances the feature in a meaningful way, as in, it's debatable if the user having a known set refresh time improves the feature, as the point is mostly to discourage attackers, not help the user by letting them know exactly when they can retry.

I didn't go that route because it required more book keeping and i wasn't sure what the final solution would be. But if your more sure, we can do that.

src/clj/pyregence/authentication.clj
drewverlee: me and katy agreed on 6, idk why that happened but this isn't a mistake. happy to make it 5 i guess idk.

src/clj/pyregence/authentication.clj
drewverlee: I don't think so because that makes it a tad harder to change and move around, and we're not that concerned about persisting this across server restarts. 

Where we would need to persist this to the DB is if we had multiple Pyrecast servers running...

src/clj/pyregence/jobs.clj
drewverlee: good call thanks

src/cljs/pyregence/pages/login.cljs
drewverlee: > join lets

done, good call

drewverlee: > email is shadowed... do

"shadowing" aka having a more local meaning to a word, is fine and i don't think lsp should warn about it.

Put another way, the issue here isn't the shadowing, its that "email" is a def ratom and not local to a component and has it's value populated from the return value from the go block/server. 

the better solution would be to have the go block return the server value including the email and funnel that into a compoment local ratom state.

src/clj/pyregence/authentication.clj
danielhvs: Yes. `loop` behaves like `let` when there is no `recur`. So it is not a real loop, just a let.

src/clj/pyregence/authentication.clj
danielhvs: Let's do the atom's approach since we have already code in place.

src/cljs/pyregence/pages/login.cljs
danielhvs: ok, no big deal, it is just the "linting error" that caught that.

danielhvs (approved):
let's deploy this and improve later if needed

PR #1136, opened 2026-07-28
PYR1-1548 touch ups

drewverlee:
This PR is to address the following requests:

> Toast for missing Uppercase letters should read: Your password must have at least one uppercase letter.” << it’s just missing the word “letter”.

> Same for lowercase letters.

Added the word letter to both messages

> Toast for Password 8 and under characters long says the following: “Your password must be at least 8 characters long.”

> Anything under 12 characters should explain what they need to have at least 12 characters.

I assume we just want a toast saying the password needs to be over 12 characters (or letters...)

> Now that we have the explainer underneath the password entry box, I am unable to scroll to the bottom of the page and I can't see the submit button when I'm still entering my password. Once both fields are filled out and validation text goes away, I’m able to submit. If possible it would be nice to change in this ticket, however, I am open to creating a separate bug ticket since it is related to the UI rendering and not the password policy enforcement itself. Just let me know.

I fixed this by adding an overflow. PLEASE test any page that this touches. ill do it myself to. thanks!!!

### DIFF
--- a/src/cljs/pyregence/pages/register.cljs
+++ b/src/cljs/pyregence/pages/register.cljs
@@ -61,7 +61,7 @@
                           "Your password must be at least 12 characters long.")
 
                         (when (< 64 (count @password))
-                          "Your password must be less then 64 characters long.")
+                          "Your password must be less than 64 characters long.")
 
                         (when-not (= @password @re-password)
                           "The passwords you have entered do not match.")
--- a/src/cljs/pyregence/pages/register.cljs
+++ b/src/cljs/pyregence/pages/register.cljs
@@ -57,8 +57,11 @@
                         (when-not (= @email @re-email)
                           "The emails you have entered do not match.")
 
-                        (when (< (count @password) 8)
-                          "Your password must be at least 8 characters long.")
+                        (when (< (count @password) 12)
+                          "Your password must be at least 12 characters long.")
+
+                        (when (< 64 (count @password))
+                          "Your password must be less then 64 characters long.")
 
                         (when-not (= @password @re-password)
                           "The passwords you have entered do not match.")
--- a/src/cljs/pyregence/pages/register.cljs
+++ b/src/cljs/pyregence/pages/register.cljs
@@ -57,12 +57,6 @@
                         (when-not (= @email @re-email)
                           "The emails you have entered do not match.")
 
-                        (when (< (count @password) 12)
-                          "Your password must be at least 12 characters long.")
-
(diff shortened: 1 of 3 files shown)


src/cljs/pyregence/pages/register.cljs
```diff
@@ -57,8 +57,11 @@
                         (when-not (= @email @re-email)
                           "The emails you have entered do not match.")
 
-                        (when (< (count @password) 8)
-                          "Your password must be at least 8 characters long.")
+                        (when (< (count @password) 12)
+                          "Your password must be at least 12 characters long.")
+
+                        (when (< 64 (count @password))
+                          "Your password must be less then 64 characters long.")
```
kbeehler-sig: @drewverlee : Minor correction.

"Your password must be less than 64 characters long."

kbeehler-sig (dismissed):

src/cljs/pyregence/pages/register.cljs
drewverlee: done

danielhvs:
## 🔴 Requested changes

`register.cljs:60-64` duplicates the length checks in `validations.cljs:14-17`, so a short password fires two toasts for the same error. `"less than 64"` is also wrong - 64 is valid (`authentication.clj:71`, `validations.cljs:16-17`).

Consolidate: have `->toast` return **all** failing messages, then splice into `errors` and delete the explicit `12`/`64` `when` blocks.

```clojure
;; validations.cljs
(defn ->toasts
  [password]
  (into []
        (comp (remove :validation/valid?)
              (map #(str "Your password must have "
                         (str/lower-case (:validation/text %)))))
        (data password)))
```

```clojure
;; register.cljs - drop the explicit 12/64 checks
(into (remove nil? [,,,])
      (password->validations/->toasts @password))
```

Same swap in `reset_password.cljs:42`.

image

danielhvs:
overflow: LGTM

image

image

image

image

drewverlee:
@danielhvs 

> register.cljs:60-64 duplicates the length checks in validations.cljs:14-17, so a short password fires two toasts for the same error. "less than 64" is also wrong - 64 is valid (authentication.clj:71, validations.cljs:16-17).

Done. thanks

> Consolidate: have ->toast return all failing messages, then splice into errors and delete the explicit 12/64 when blocks.

That's not what the ticket or designs ask for so lets do that in another ticket.

danielhvs (approved):
Deal!

PR #1139, opened 2026-08-06
PYR1-1568 Treat a dead cookie as anonymous on add-new-user and get-fire-names

sifbiri:
## Description
After you log out, two pages still believed your old cookie.

Since #1134 most routes check a session is alive before trusting it, but these two can't.
`add-new-user` also backs the public Register page and `get-fire-names` feeds the public map, so
neither can require a login. A logged-out admin could still add a user to an organization, and a
logged-out cookie could still see that person's match-drop names. Both now check first and treat a
dead cookie as anonymous, which each page already handles.

Tests cover logging out and timing out. Removing either guard, or narrowing `live?` to timeouts
only, fails the suite.

`bb test`, `bb test-clj` and clj-kondo are green. Also checked over HTTP: a dead admin cookie
posting `{:org-id 2}` gets a 403, the same cookie live gets a 200, and signup is unaffected.

### DIFF
--- a/src/clj/pyregence/session.clj
+++ b/src/clj/pyregence/session.clj
@@ -8,7 +8,7 @@
 (def ^:private default-idle-timeout-min     15)  ; NIST 800-63B AAL3 / PCI DSS 8.2.8
 (def ^:private default-absolute-timeout-min 420) ; 7 h
 
-(defn expired?
+(defn- expired?
   "Fail-closed: a session missing either timestamp counts as expired; an unauthenticated one never expires."
   [{:keys [user-id created-at last-active]} now idle-ms absolute-ms]
   (boolean (when user-id
@@ -28,7 +28,7 @@
             (timeout-ms :pyregence.auth/idle-timeout-min     default-idle-timeout-min)
             (timeout-ms :pyregence.auth/absolute-timeout-min default-absolute-timeout-min)))
 
-(defn invalidated?
+(defn- invalidated?
   "Created strictly before the user's invalidation point (set on logout / newer login).
    Strict `<` so a fresh login at the same instant survives; invalidated-at 0 = never."
   [{:keys [user-id created-at]} invalidated-at]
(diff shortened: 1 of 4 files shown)


src/clj/pyregence/session.clj
```diff
@@ -0,0 +1,95 @@
...
+  "Session liveness. Sessions are stateless signed cookies, so one is dead when its stamps
+   fall outside the configured timeouts or it predates the user's invalidation cutoff.
+   `live?` is what a public route asks, since the auth gate cannot enforce liveness for it."
+  (:require [triangulum.config   :refer [get-config]]
+            [triangulum.database :refer [call-sql]]))
+
+(def ^:private default-idle-timeout-min     15)  ; NIST 800-63B AAL3 / PCI DSS 8.2.8
+(def ^:private default-absolute-timeout-min 420) ; 7 h
+
+(defn expired?
```
danielhvs:
```suggestion
(defn- expired?
```

src/clj/pyregence/session.clj
```diff
@@ -0,0 +1,95 @@
...
+  (* 60000 (or (get-config config-key) default-min)))
+
+(defn timed-out?
+  "Whether the session is past its configured idle or absolute timeout as of `now`."
+  [session now]
+  (expired? session now
+            (timeout-ms :pyregence.auth/idle-timeout-min     default-idle-timeout-min)
+            (timeout-ms :pyregence.auth/absolute-timeout-min default-absolute-timeout-min)))
+
+(defn invalidated?
```
danielhvs:
```suggestion
(defn- invalidated?
```

danielhvs (approved):
Reviewed and analyzed this on my own as well as in collaboration with Claude Code.

I'd suggest `1.` more (to improve performace). `2.` I'd say is more a refactor for maintainability:

1. `capabilities.clj` - avoid a DB round trip on the public map for most users

`session/live?` now issues a `get_user_session_invalidated_at` query on every logged-in map load,
but the session only changes the output when `match-drop-access?` is set. Reordering keeps the
query rare without changing semantics:

```clojure
(let [{:keys [user-id match-drop-access?]} session
      match-drop-names (when (and match-drop-access? (session/live? session))
                         ,,,)]
```

2. session.clj - the hardcoded result key can fail open

(:get_user_session_invalidated_at (first ...)) is correct today (triangulum builds rows with
as-unqualified-lower-maps, so the column is the function name). But if that name ever changes,
the lookup returns nil, (or ... 0) turns it into "never invalidated", and dead cookies are
accepted again, silently. Dropping sql-primitive was the right call since (comp val first first)
throws on zero rows, but this keeps both properties:

`  (or (some-> (call-sql "get_user_session_invalidated_at" user-id) ffirst val) 0)`

